PR #4, opened 2015-07-14
Simplify configuration

ppikula:
Sorry for putting everything in one PR :( - these are changes made for the euc tutorial.


lavrin:
I've added a number of commits, specifically enabling to use Amoc on Ubuntu Precise 64 (Vagrant box ubuntu/precise64). bdb805e might carry incompatibilities - could you recheck in your original env whether everything still works?

PR #14, opened 2016-04-13
Make number of repetitions configurable

chrzaszcz:
Additional fixes:
- fix deps to work with R18 and accept later versions of escalus
- clean relx setup

### DIFF
--- a/src/amoc_user.erl
+++ b/src/amoc_user.erl
@@ -9,6 +9,10 @@
 -define(REPEAT_INTERVAL,
     application:get_env(amoc, repeat_interval, ?REPEAT_INTERVAL_DEFAULT)).
 
+-define(REPEAT_NUM_DEFAULT, infinity).
+-define(REPEAT_NUM,
+    application:get_env(amoc, repeat_num, ?REPEAT_NUM_DEFAULT)).
+
 %% API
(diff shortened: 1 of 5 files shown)


src/amoc_user.erl
```diff
@@ -9,6 +9,10 @@
 -define(REPEAT_INTERVAL,
     application:get_env(amoc, repeat_interval, ?REPEAT_INTERVAL_DEFAULT)).
 
+-define(REPEAT_NUM_DEFAULT, infinity).
+-define(REPEAT_NUM,
+    application:get_env(amoc, repeat_num, ?REPEAT_NUM_DEFAULT)).
```
studzien: Wouldn't a function definition (e.g. ``repeat_num/0``) be better for this purpose?
Yes, the repeat interval value is defined with macros as well, but I believe it also should be changed.

studzien: See: https://medium.com/@erszcz/when-not-to-use-macros-in-erlang-1d3f10d377f

chrzaszcz: I will kill this monstrosity with fire.

studzien: 😎

PR #16, opened 2016-05-02
[WIP] Configuration helpers

studzien:
The ``amoc_config`` helper (wrapper around ``application:get/set_env``) is added to facilitate usage of dynamic configuration inside the scenarios.
Also, mainly for usage with Docker containers, all environmental variables starting with ``AMOC_name`` will be stored as application variables (the value of these variables will be parsed as Erlang terms).

To do before this PR might be considered in a ready-to-merge state:
- [ ] update README
- [ ] prepare a scenario using dynamic config
- [ ] fix the current mess with the calls to ``application:get_env`` inside macros.

### DIFF
--- a/src/amoc_config.erl
+++ b/src/amoc_config.erl
@@ -1,0 +1,67 @@
+%%==============================================================================
+%% Copyright 2016 Erlang Solutions Ltd.
+%% Licensed under the Apache License, Version 2.0 (see LICENSE file)
+%%==============================================================================
+-module(amoc_config).
+
+-export([start/0,
+         set/2,
+         get/1,
+         get/2,
+         fetch/1]).
+
+%% For tests
+-export([set_env_variables/2]).
+
+%% ------------------------------------------------------------------
+%% API
+%% ------------------------------------------------------------------
+-spec start() -> ok.
+start() ->
+    ok = set_env_variables(os:getenv()).
+
+-spec set(atom(), any()) -> ok.
+set(Name, Value) ->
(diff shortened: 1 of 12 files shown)


src/amoc_config.erl
```diff
@@ -0,0 +1,67 @@
...
+set_env_variables(EnvVars, Filter) ->
+    Vars = parse_variables(EnvVars),
+    [ ok = set(Name, Value) || {Name, Value} <- Vars, Filter(Name) ],
+    ok.
+
+-spec parse_variables([string()]) -> [{atom(), any()}].
+parse_variables(EnvVars) ->
+    Vars = [ re:split(Var, "=", [{return, list}, {parts, 2}])
+             || Var <- EnvVars ],
+    %% list_to_atom should be safe here, as we are in control of env variables
```
michalwski: How about putting all possible params to the `app.config` or `sys.config` with their defaults. Thanks to that:

1. We can use list_to_existing_atom
2. We have well defined default values.

Does it make sense?

studzien: If we want to use a custom setting (like list of xmpp servers or whatever that makes sense only inside a scenario) then we would need to define it there as well.
The goal of introducing this is to allow Docker container configuration done only via environmental variables.
Let's start with that, that I'm not sure if this is even a good approach :)

studzien: We've agreed that environmental variables should be applied to the configuration in the Docker init script, before amoc starts as a service.

studzien:
overriden by #32

PR #18, opened 2016-07-25
AMOC HTTP REST API handling status

kklocek:
AMOC status handler. We can retrieve status of running AMOC by `GET /status`.


src/amoc_api_status_handler.erl
```diff
@@ -9,20 +9,17 @@
...
--type state() :: #state{}.
-
--spec trails() -> [tuple()] .
+-spec trails() -> trails:trails().
 trails() ->
     Metadata =
         #{get =>
           #{tags => ["status"],
-            description => "Gets AMOC status, whether it is running or not",
+            description => "Gets AMOC status, whether it is running or not. Returns {\"status\": true | false}",
```
elvisci: Line 22 is too long:             description => "Gets AMOC status, whether it is running or not. Returns {\"status\": true | false}",.

test/amoc_api_status_handler_SUITE.erl
```diff
@@ -4,260 +4,76 @@
...
-         test_list/1,
-         test_start_fail_no_scenario/1,
-         test_start_success/1,
-         test_load_good/1,
-         test_load_compilation_err/1,
-         test_ping_nodes/1,
-         test_get_test_result/1,
-         test_get_tests_results/1,
-         test_get_wrong_test_result/1]).
+-export([works_when_amoc_running_offline/1,
```
mentels: The test cases' names are not too good. Please refer to our rules to make them better.

For example: `get_status_returns_false_when_amoc_down`.

mentels:
Refer to the our rules and fix the commit message accordingly.

test/amoc_api_status_handler_SUITE.erl
```diff
@@ -4,260 +4,76 @@
...
-           ],
-    meck:new(amoc_dist, [unstick]),
-    Fun = fun(_,1,_) -> ok end,
-    meck:expect(amoc_dist, do, Fun),
     %% when
-    Result = amoc_api_scenario_handler:process_json(Term,#state{action = start}),
-    meck:wait(amoc_dist, do, ['mongoose_simple', 1, 10], 1000),
-    meck:unload(amoc_dist),
-    {CodeHttp,JSON} = Result,
+    Result = httpc:request(get, {get_url() ++  "/status", [{"Accept", "application/json"}]}, [], []),
```
mentels: The whole WHEN should be hidden in a function like: `{HttpCode, JsonBody} = send_get(...)`.

To much details in the test function.

elbrujohalcon: Can I add that you should not use `httpc`, in general? For tests, it's ok, but since you're using this project to learn, I think you can take this opportunity to start using [fusco](https://github.com/esl/fusco) as an http client.

PR #25, opened 2016-07-29
Simple api2 test status

kklocek:
Checking test status via `GET /scenarios/:id`.


src/amoc_api_test_handler.erl
```diff
@@ -11,34 +11,34 @@
...
-        {[tuple()], cowboy:req(), state()}.
+-spec content_types_provided(cowboy_req:req(), state()) -> 
+        {[tuple()], cowboy_req:req(), state()}.
 content_types_provided(Req, State) ->
     {[
         {<<"application/json">>, to_json}
     ], Req, State}.
 
--spec from_json(cowboy:req(), state()) -> {halt | false, cowboy:req(), state()}.
+-spec from_json(cowboy_req:req(), state()) -> {halt | false, cowboy_req:req(), state()}.
```
elvisci: Line 41 is too long: -spec from_json(cowboy_req:req(), state()) -> {halt | false, cowboy_req:req(), state()}..

src/amoc_api_status_handler.erl
```diff
@@ -26,26 +26,26 @@ trails() ->
     },
     [trails:trail("/status", amoc_api_status_handler, [], Metadata)].
 
--spec init(tuple(), cowboy:req(), state()) -> {upgrade, protocol, cowboy_rest}.
+-spec init(tuple(), cowboy_req:req(), state()) -> {upgrade, protocol, cowboy_rest}.
```
elvisci: Line 29 is too long: -spec init(tuple(), cowboy_req:req(), state()) -> {upgrade, protocol, cowboy_rest}..

src/amoc_api_test_handler.erl
```diff
@@ -11,34 +11,34 @@
 
 -type state() :: list().
 
--spec init(tuple(), cowboy:req(), state()) -> {upgrade, protocol, cowboy_rest}.
+-spec init(tuple(), cowboy_req:req(), state()) -> {upgrade, protocol, cowboy_rest}.
```
elvisci: Line 14 is too long: -spec init(tuple(), cowboy_req:req(), state()) -> {upgrade, protocol, cowboy_rest}..

src/amoc_api_scenario_handler.erl
```diff
@@ -81,8 +81,8 @@ resource_exists(Req, State = #state{resource = Resource}) ->
 -spec to_json(cowboy_req:req(), state()) -> 
     {jsx:json_text(), cowboy_req:req(), state()}.
 to_json(Req0, State = #state{resource = Resource}) ->
-    %% Need to retrive state of module here
-    Reply = jsx:encode([{module, Resource}]),
+    Status = amoc_controller:test_status(erlang:list_to_atom(Resource)),
```
mentels: What are possible test status values? It seems that `amoc_controller:test_status/1` returns true or false. It's not logical to have test status of true or false regarding a test scenario.

As far as I remember a test can be either running or loaded from the Amoc perspective. And it would be nice if these statuses have a common set with those defined on the Tide side. @kklocek and @arkgil can you discuss that together an describe possible status values somewhere? Then @kklocek please adjust this PR.

Additionally I want to have this status values described within the spec by something like:

```erlang

-type scenario_status() :: loaded | running.

 -spec test_status(scenario_name()) -> scenario_status().
test_status(ScenarioName) ->
    {ok, Res} = gen_server:call(?SERVER, {status, ScenarioName}),
    Res.
```

test/amoc_api_scenario_handler_SUITE.erl
```diff
@@ -51,6 +57,7 @@ end_per_testcase(_, _Config) ->
 
 get_scenario_status_returns_200_when_scenario_exists(_Config) ->
     %% given
+    given_amoc_controller_prepared(true),
```
mentels: Literally passing atom `true` here is a bad style. When you look at this code you have no idea what it means.

test/amoc_api_scenario_handler_SUITE.erl
```diff
@@ -82,7 +125,7 @@ patch_scenario_returns_400_when_malformed_request(_Config) ->
     URL = get_url() ++ "/scenarios/sample_test",
     RequestBody = jsx:encode([{bad_key, bad_value}]),
     %% when
-    {CodeHttp, Body} = patch_request(URL, RequestBody),
+    {CodeHttp, _Body} = patch_request(URL, RequestBody),
     %% then
     %% Maybe check Body, as answer format will be ready
     ?assertEqual(400, CodeHttp).
```
mentels: The test says `404` but you're checking `400`.

kklocek: Everything is ok, there are 2 other tests, see full file.

src/amoc_api_scenario_handler.erl
kklocek: Fixed.

Possible scenario status values: loaded, running or finished.

src/amoc_api_status_handler.erl
```diff
@@ -26,26 +26,27 @@ trails() ->
...
 content_types_provided(Req, State) ->
     {[{<<"application/json">>, to_json}], Req, State}.
 
--spec to_json(cowboy:req(), state()) ->
-    {binary(), cowboy:req(), state()}.
+-spec to_json(cowboy_req:req(), state()) ->
+    {binary(), cowboy_req:req(), state()}.
 to_json(Req0, State) ->
     Status = get_status(),
     StatusJson = jsx:encode([{result, Status}]),
```
mentels: Let's change result into something more meaningful: `scenario_status` maybe?

PR #26, opened 2016-08-01
Better REST API status description

kklocek:
When we do `GET /status` on response we have `{ node_status: up | down }` instead of `{ status: true | false }`.


arkgil:
@mentels this also looks fine to me.

PR #27, opened 2016-08-02
Add node listing on GET /nodes

kklocek:
Now when we do `GET /nodes` we invoke function `amoc_dist:ping_nodes()`, and on response we get:
`
{
    "nodes":
    {
        nodeName1: "up" | "down",
        nodeName2:  "up" | "down"
        ...
}
`


arkgil:
@mentels PR looks good to me, tests are also passing.

PR #30, opened 2016-08-02
Change way of getting scenario status

kklocek:
Now if you want to get scenario status, yo will get status running on the whole
cluster, previously you get information from node.

### DIFF
--- a/src/amoc_api_scenario_handler.erl
+++ b/src/amoc_api_scenario_handler.erl
@@ -81,8 +81,7 @@ resource_exists(Req, State = #state{resource = Resource}) ->
 -spec to_json(cowboy_req:req(), state()) -> 
     {jsx:json_text(), cowboy_req:req(), state()}.
 to_json(Req0, State = #state{resource = Resource}) ->
-    Status = amoc_controller:test_status(
-                     erlang:list_to_atom(Resource)),
+    Status = amoc_dist:test_status(erlang:list_to_atom(Resource)),
     Reply = jsx:encode([{scenario_status, Status}]),
     {Reply, Req0, State}.
 
--- a/src/amoc_controller.erl
+++ b/src/amoc_controller.erl
@@ -116,7 +116,7 @@ handle_call(users, _From, State) ->
              {last, ets:last(?TABLE)}],
     {reply, {ok, Reply}, State};
 handle_call({status, Scenario}, _From, State) ->
-    Res = case does_scenario_exists(Scenario) of
+    Res = case does_scenario_exist(Scenario) of
         true -> check_test(Scenario, State#state.scenario);
         false -> error
     end,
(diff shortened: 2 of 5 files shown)


src/amoc_dist.erl
```diff
@@ -95,6 +102,25 @@ ping_node(Node) ->
...
+
+-spec pick_status(atom(), [amoc_controller:scenario_status()]) ->
+                                amoc_controller:scenario_status().
+pick_status(StatusList, [H | T]) ->
+    case lists:member(H, StatusList) of
+        true -> H;
+        false -> pick_status(StatusList, T)
+    end.
+
+-spec get_node_test_status(atom(), atom()) ->
```
bszaf: Maybe change spec to (amoc:scenario(), node()).

src/amoc_dist.erl
```diff
@@ -95,6 +102,25 @@ ping_node(Node) ->
         pang ->
               pang
     end.
+
+-spec pick_status(atom(), [amoc_controller:scenario_status()]) ->
```
bszaf: Maybe change atom() to amoc_controller:scenario_status().

src/amoc_dist.erl
```diff
@@ -31,6 +32,12 @@ ping_nodes() ->
 do(Scenario, Start, End) ->
     do(Scenario, Start, End, []).
 
+-spec test_status(atom()) -> amoc_controller:scenario_status().
```
bszaf: Maybe change atom() to amoc:scenario().

src/amoc_dist.erl
```diff
@@ -31,6 +32,12 @@ ping_nodes() ->
...
+test_status(ScenarioName) ->
```
bszaf: When calling function with non-existing scenario as parameter it returns 'loaded'.

src/amoc_controller.erl
```diff
@@ -292,3 +293,19 @@ get_test_status() ->
         [] -> finished;
         _Children -> running
     end.
+
+-spec does_scenario_exists(atom()) -> boolean().
+does_scenario_exists(Scenario) ->
```
michalwski: A typo here? I think the correct name should be `does_scenario_exist` without the ending `s`.

src/amoc_api_scenario_handler.erl
```diff
@@ -81,8 +81,7 @@ resource_exists(Req, State = #state{resource = Resource}) ->
 -spec to_json(cowboy_req:req(), state()) -> 
     {jsx:json_text(), cowboy_req:req(), state()}.
 to_json(Req0, State = #state{resource = Resource}) ->
-    Status = amoc_controller:test_status(
-                     erlang:list_to_atom(Resource)),
+    Status = amoc_dist:test_status(erlang:list_to_atom(Resource)),
```
michalwski: Can we assume that in correct use case the target atom exists? If so, I suggest using function `list_to_existing_atom`. It's not safe to covert to atom from user's input. Atoms are not garbage collected so it's fairly easy to misuse this API and crash the app.

Moreover I've noticed that in first place there is conversion from binary to list. I think there is no need for that because there is `binary_to_atom/2` function which could be used here.

kklocek: Ad 1. It's safe here because when we want to do a `GET` request cowboy checks whether scenario exists on filesystem, so in this line we will get a name of a existing module (but maybe not loaded)
Ad 2. We use `binary_to_list/2` in `rest_init` function to check later in `resource_exists` whether our scenario is on disk. Without that we would need to convert our binary/atom to list, so in my opinion it's not a problem.

michalwski: I get it, thanks for the explanation :)

PR #31, opened 2016-08-03
Include scenario sources in release

bszaf:
To be able to show users source of scenario, we should include them in release. Directory "scenarios" is also used for storage scenarios uploaded by user.

### DIFF
--- a/relx.config
+++ b/relx.config
@@ -25,3 +25,7 @@
 {sys_config, "priv/app.config"}.
 {include_src, false}.
 {vm_args, "./priv/vm.args"}.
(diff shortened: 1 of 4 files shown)


relx.config
```diff
@@ -25,3 +25,7 @@
 {sys_config, "priv/app.config"}.
 {include_src, false}.
 {vm_args, "./priv/vm.args"}.
+{overlay, [
+           {mkdir, "ebin"},
```
arkgil: `ebin/` directory name is misleading, could you please change it to more meaningful?

PR #35, opened 2016-08-16
Simple api2 run scenario fix

kklocek:
Previously we have some bugs and problems with running uploaded scenario:
* there was no information that scenario is not started
* scenario crashes if it uses lager which need special compile flag during compilation
* if we compiled scenario then we couldn't load it because in one line we provide directory name as atom, not a string and we couldn't use scenario later.

### DIFF
--- a/src/amoc.app.src
+++ b/src/amoc.app.src
@@ -27,12 +27,5 @@
           {[amoc, counters, connections], spiral, []},
           {[amoc, counters, connection_failures], spiral, []}
          ]}
-  ]},
-  {lager, [
-  {handlers, [
-    {lager_console_backend, info},
-    {lager_file_backend, [{file, "error.log"}, {level, error}]},
-    {lager_file_backend, [{file, "console.log"}, {level, info}]}
-  ]}
   ]}
  ]}.
--- a/src/amoc.app.src
+++ b/src/amoc.app.src
@@ -7,6 +7,7 @@
                   kernel,
                   stdlib,
                   lager,
+                  jiffy,
                   exometer,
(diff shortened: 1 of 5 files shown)


src/amoc_api_scenario_handler.erl
```diff
@@ -173,3 +175,14 @@ get_users_from_body(Req) ->
               {error, bad_request, Req2}
     end.
 
+-spec get_result([ok | {error, term()}]) -> started | error.
+get_result(Result) ->
+    Res = lists:all(fun(X) -> X == ok end, Result),
+    case Res of
+        true -> started;
+        false ->
+                %% Internal server error, log problems
```
arkgil: What does 'log problems' refer to?

arkgil:
Looks good to me.

michalwski:
Could you please describe the bug shortly in the PR desc?

src/amoc.app.src
```diff
@@ -7,6 +7,7 @@
                   kernel,
                   stdlib,
                   lager,
+                  jiffy,
```
michalwski: I know we need `jiffy` because of `trails`. We also have `jsx` which does pretty much the same as `jsx`. Could you check if we can replace `jsx` with `jiffy`? If so, then we should do this in a separate PR.

kklocek:
Added bugs description.

src/amoc.app.src
kklocek: We also need a `jiffy` for `cowboy-swagger` .
So, the question is to replace `jsx` with `jiffy`? I will try.

PR #36, opened 2016-08-17
Replace jsx with jiffy in amoc

kklocek:
Previously we had jiffy and jsx. We don't need two JSON parsers, and we need to have jiffy, so now we use only jiffy in code.


arkgil:
These changes look fine to me. @michalwski can you make final review?

PR #38, opened 2016-08-29
Replace httpc with fusco

kklocek:
Instead `httpc` we want to use `fusco` only for testing purpose.
We need it also because on Travis CI we have OTP 17.5 where `httpc` do not support `PATCH` requests.

### DIFF
--- a/scenarios/mongoose_pubsub_docker.erl
+++ b/scenarios/mongoose_pubsub_docker.erl
@@ -50,7 +50,18 @@ start(MyId) ->
 
 work_as_publisher(MyId, Client) ->
     create_pubsub_node(Client, pubsub_node(MyId)),
-    publish_items_forever(Client, MyId, pubsub_node(MyId), 1).
+    case MyId rem 3 of
+        0 -> publish_many_items(Client, MyId, pubsub_node(MyId), 1, 20000);
+        _ -> publish_items_forever(Client, MyId, pubsub_node(MyId), 1)
+    end.
+
+publish_many_items(_, _, _, _, 0) -> ok;
+
+publish_many_items(Client, MyId, Node, ItemId, Count) ->
+    timer:sleep(?DELAY_BETWEEN_MESSAGES),
+    lager:debug("Publisher ~p publishing item ~p.", [MyId, ItemId]),
+    publish(Client, integer_to_binary(ItemId), Node),
+    publish_many_items(Client, MyId, Node, ItemId + 1, Count - 1).
 
 publish_items_forever(Client, MyId, Node, ItemId) ->
     timer:sleep(?DELAY_BETWEEN_MESSAGES),
--- a/src/amoc.app.src
+++ b/src/amoc.app.src
@@ -13,7 +13,8 @@
                   ssl,
                   trails,
                   cowboy,
-                  cowboy_swagger
+                  cowboy_swagger,
+                  fusco
                  ]},
   {mod, { amoc_app, []}},
   {env, [{repeat_interval, 60000},
--- a/src/amoc.app.src
+++ b/src/amoc.app.src
@@ -13,8 +13,7 @@
                   ssl,
                   trails,
                   cowboy,
-                  cowboy_swagger,
-                  fusco
+                  cowboy_swagger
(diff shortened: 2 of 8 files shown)


kklocek:
Also I fix dialyzer error with infinite loop in `scenarios/mongoose_pubsub.docker.erl` .

src/amoc.app.src
```diff
@@ -13,7 +13,8 @@
                   ssl,
                   trails,
                   cowboy,
-                  cowboy_swagger
+                  cowboy_swagger,
+                  fusco
```
mentels: If fusco is only used in the tests it doesn't have to be included in the app.src file.

test/amoc_api_scenario_handler_SUITE.erl
```diff
@@ -177,45 +170,37 @@ get_url() ->
...
-    Type = "application/json",
-    HTTPOpts = [],
-    Opts = [],
-    Result = httpc:request(patch,
-                           {URL, Header, Type, RequestBody},
-                           HTTPOpts,
-                           Opts),
-    {ok, {{_HttpVsn, CodeHttp, _Status},_, Body}} = Result, 
+patch_request(Path, RequestBody) ->
+    request(<<"PATCH">>, Path, RequestBody).
```
mentels: Why here the HTTP Verb is passed as binary while in the line 173 as atom? That should be unified.

test/amoc_api_scenarios_handler_SUITE.erl
```diff
@@ -126,40 +121,31 @@ get_url() ->
...
-    Header = "",
-    Type = "application/json",
-    HTTPOpts = [],
-    Opts = [],
-    Result = httpc:request(post,
-                           {URL, Header, Type, RequestBody},
-                           HTTPOpts,
-                           Opts),
-    {ok, {{_HttpVsn, CodeHttp, _Status},_, Body}} = Result, 
+post_request(Path, RequestBody) ->
```
mentels: As the requests are used in different test modules they should be extracted into a separate module with test utilities.

scenarios/mongoose_pubsub_docker.erl
```diff
@@ -50,7 +50,18 @@ start(MyId) ->
 
 work_as_publisher(MyId, Client) ->
     create_pubsub_node(Client, pubsub_node(MyId)),
```
mentels: @kklocek I guess the solution is based on the fact that `MyId rem 3 /= 0` always holds true - am I right? If so, then is should be documented and explained in the commit message.

Apart from that, I'm not sure if this scenario is still valid after this change - probably should be checked by @chrzaszcz.

Anyway, I suggest moving this piece of code onto separate branch and creating a new PR for it describing exactly which dialyzer error it addresses. I'm not sure that you picked a right solution.

kklocek:
After that, please review/merge #39 because it contains good solution for dialyzer errors.

test/amoc_api_helper.erl
```diff
@@ -0,0 +1,54 @@
+-module(amoc_api_helper).
```
mentels: Beautiful module :) However, one thing is missing. In line 39 you start fusco which gives you a client, for each request. [fusco:start/2](https://github.com/esl/fusco/blob/master/src/fusco.erl#L77) create a new gen_server process and it is not killed after the request is done - which is bad because if someone make use of the utils module and his or her test SUITE will generate 10000000000 request the he or she will end up with many orphaned processes. There seem to be 2 solutions:

* either add an API for obtaining/releasing (get_client/0, release_client/1) the HTTP client and then pass it on the the requests, or
* disconnect the client after each request.

Think which one is better and choose @kklocek !
*

scenarios/mongoose_pubsub_docker.erl
kklocek: Fixed in #39

test/amoc_api_helper.erl
kklocek: Fixed in new commit.
Now we disconnect client after each request with `fusco:disconnect/1`.

PR #41, opened 2016-09-26
Use amoc_config:get for getting setting instead of os:get_env

bszaf:


scenarios/mongoose_simple_docker.erl
```diff
@@ -163,21 +163,12 @@ make_jid(Id) ->
...
-    list_to_tuple(List).
-
 -spec set_env_interarrival() -> ok.
 set_env_interarrival() ->
-    Interarrival = case os:getenv("AMOC_INTERARRIVAL") of
-                       false -> 100;
-                       List -> list_to_integer(List)
-                   end,
+    Interarrival = amoc_config:get('INTERARRIVAL'),
     application:set_env(amoc, interarrival, Interarrival).
```
mentels: This is bad, because we break the abstraction and set `interarrival` instead of having it in the amoc_config:get/1 directly by passing `AMOC_interarrival` instead of `AMOC_INTERARRIVAL`. It has to be fixed.

bszaf: Fixed. Now we have to pass directly `AMOC_interarrival`

PR #43, opened 2016-11-08
Add code:purge to scenarios API handler and make console target

kklocek:
During REST API development I forgot to call `code:purge/1` before loading new scenario.

Also I add new makefile target for console.

### DIFF
--- a/makefile
+++ b/makefile
@@ -40,6 +40,9 @@ eunit:
 
 test: compile eunit ct prop
 
+console:
+	erl -pa ebin/ -pa deps/*/ebin
+
 dialyzer/erlang.plt:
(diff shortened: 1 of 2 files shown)


makefile
```diff
@@ -40,6 +40,9 @@ eunit:
 
 test: compile eunit ct prop
 
+console:
+	erl -pa ebin/ -pa deps/*/ebin
```
michalwski: Where scenarios are compiled to? `ebin` or `scenario` dir?

kklocek: To `ebin` directory. I just add this target for testing purpose, I use it multiple times while testing amoc.

Issue #45, opened 2016-11-14
copy scenarios to scenarios/


bartekgorny:
Every time "make rel" is called, a directory with scenarios is copied to "scenarios/" - which means that you get "_rel/scenarios/scenarios/scenarios" repeated as many times as you have ran "make rel". All this goes to slaves, and seems to result in slaves running an old versions of scenarios.

kklocek:
@bartekgorny Thanks for that! It's somewhat strange, I made a PR which in my case fixes it.

Please test it if you will have more time.

studzien:
Resolved by #46

PR #46, opened 2016-11-15
Fix multiple scenarios dir generation on making release:

kklocek:
It solves probably #45 . Earlier we have that line in `relx.config` because we need to include `scenarios` dir when making a release, however this attempt also works.


berespt:
It fixes the #45 problem for me too. When can we expect it to be merged?

PR #49, opened 2016-11-15
Support for erlang otp 19

ludwikbukowski:


arkgil:
Has escalus API been changed in version 3.0.0? If that is a case, we should probably update scenarios.

ludwikbukowski:
good catch! It shouldn't but I will check

kklocek:
Any updates on that @ludwikbukowski ?

ludwikbukowski:
only one thing that breaks backward compability in Escalus is this issue https://github.com/esl/escalus/pull/117 . I think it does not affect the scenarios

PR #50, opened 2016-12-12
Switch to rebar3 and ditch previous toolset

ppikula:
Hi guys! 

Apart from tool changes I tried to not touch existing logic: ex. relx config and the makefile commands.

To give more background of this work. I wanted to have a possibility to write scenarios in Elixir, so I though about taking https://github.com/barrel-db/rebar3_elixir_compile to achieve that, but right now I'm not sure if it will be the best approach:

- I'd a bit cumbersome to pass `Elixir.sthsthst` to the amoc `do` function.
- It seems that plugin creates a strong dependency to Elixir, but there still true men out there that use only pure Erlang ;) I can see that ejabberd did some conditional compile paths, but I don't really like it
- we are not able to start elixir shell(right?), so even if we create wrappers for `do` calls we won't be able to call it 

My question is, have you thought about it? I'm not a world class elixir expert, so I would be happy to have some guidance here. I can see 3 options: 

#### 1. rebar3_elixir_compile with crazy ifs 
drawbacks above

#### 2. Throw `mix` into the mix
Pros:
- We use mix tasks when we want to swim in elixir soup. Have iex, elixir scenarios etc. Real men can still use rebar. 

Cons:

- We need to keep the some options in two places - like Erlang deps.
- We are not able to extend amoc core using elixir, because we might break the rebar workflow.

#### 3. Create a new repo that will fetch amoc as dependency

We won't break the `main` amoc, but it will be just an elixir wrapper and using ansible toys might be a bit harder.

Do you guys see any other options? Which one do you prefer? 

### Conclusion
All in all this is a topic for another pull requests, I won't make this one bigger. If you find this useful please accept, if not I can always have my own my fork and change... ;)


michalwski:
Cool! Thanks a lot!

Re writing scenario is Elixir, passing 'Elixr.sthsth' to amoc `do` is not that cumbersome, at lest for me. I can agree that it can be for someone who's writing Elixir on daily basis.

ppikula:
Ah, don't merge it. I forgot about updating the ansible files with `_rel` -> `_build `  will follow up shortly

ppikula:
Ready, I also tested the docker image and ansible jobs on 2 hosts setup - seems to work :).

rebar.config
```diff
@@ -1,26 +1,59 @@
...
-            {mochijson2, ".*", {git, "git://github.com/bjnortier/mochijson2.git", {branch, "master"}}},
-            {proper, ".*", {git, "git://github.com/manopapad/proper.git", {branch, "master"}}},
-            {recon, ".*", {git, "https://github.com/ferd/recon.git", "2.2.1"}},
-            {cowboy, ".*", {git, "https://github.com/ninenines/cowboy.git", "1.0.4"}},
-            {fusco, ".*", {git, "https://github.com/esl/fusco.git"}}
-           ]}.
+{require_otp_vsn, "^1[789]"}.
 
+{deps, [
+        {lager, "3.2.1"},
```
michalwski: I know this format is simpler and more readable. It happens form time to time that amoc is used a dep in project which is still compiled by rebar2. Could you get back to the longer dep spec compatible with rebar2?

ppikula: Hi @michalwski ,
It is more about taking deps from the hex.pm instead of github. It'd be good to tag the current version(and actually introducing tagging, at the moment I can see none, apart from the magic 0.9 version, but it is not tagged), before pushing some breaking changes in the future. OFC If you insist I can restore github deps. Additionally proper also have to be taken back from the test profile, assuming that the profiles are not supported in r2

It is good that you mentioned the rebar2 compiled project. What is a strategy for developing custom scenarios that need to be closed sources? We are going to have some and putting them in the scenarios dir makes back porting harder. We need to be careful to commit only propertiary stuff, then "open" ones and use cherry-pick to move it to your repo. How do you handle it? You said that you used is as a dep, does it help with the custom scenarios case? 

The biggest drawback of "depping" I can see is that we will lose access to the make deploy commands. On the other hand maybe it is not a best idea to keep ansible stuff as a part of amoc repo. How do you solve it?

The other option I can see is renaming `scenarios` to `examples` or  `sample_scenarios` thanks to that we can clone our custom scenarios dir without touching core of amoc. Makefile could fallback to copy examples to the scenarios dir when the scenarios dir is not present, just to keep things going by default. It still needs a manual work(clone scenario repo), but It is better that current situation. 

All in all if you know ideal approach let me know!

To advertise ditching rebar2 compatibility - developer preview :) : https://github.com/ppikula/amoc_templates BTW I'm open to transfer ownership of this plugin to ESL when it is done. I'm not interested in maintaining it.

ppikula: One more thing I found this in one of your deps, which is rebar3 one:

https://github.com/inaka/cowboy-trails/blob/master/rebar.config.script

michalwski:
>It is more about taking deps from the hex.pm instead of github. It'd be good to tag the current version(and actually introducing tagging, at the moment I can see none, apart from the magic 0.9 version, but it is not tagged), before pushing some breaking changes in the future.

Good point :) We'll do this

>It is good that you mentioned the rebar2 compiled project. What is a strategy for developing custom scenarios that need to be closed sources? We are going to have some and putting them in the scenarios dir makes back porting harder. We need to be careful to commit only propertiary stuff, then "open" ones and use cherry-pick to move it to your repo. How do you handle it? You said that you used is as a dep, does it help with the custom scenarios case?

In our CLT (continuous load testing) infrastructure (for MongooseIM) [1] we have an amoc node running with REST API [2] enabled. All custom scenarios (not included in amoc repo) are started by amoc's REST API. I don't think this is ideal solution but it works for us. Also in CLT custom scenarios are not closed, but only because we decided to make them open and possible to download.

Regarding the deps compatibility with rebar2 - I'm convinced, let's push rebar3-cation further!

Hope this helps.

[1]. http://tide.erlang-solutions.com/public
[2]. https://github.com/esl/amoc/blob/a6cee04c024d9ed92f4894a9e819eb500ba00b48/REST_API_DOCS.md

studzien:
While the HTTP API for deploying scenarios is sufficient in case of XMPP/HTTP scenarios I'm afraid that it won't work in case the scenario itself is using a dependency that needs to be fetched.
Therefore, I think that in the long run we won't get away with making amoc truly "deppable".

Personally I really like the idea of decoupling the ansible part.
@ppikula did you have any thoughts about this approach?

I was thinking about providing the deployed tarball path to `make deploy` command.
On the other hand, currently the `ansible` directory contains all the configuration templates. Perhaps the automation part should only append the cluster-specific configuration and leave everything else as it is in the tarball.

What do you think?

rebar.config
bryanhuntesl: @michalslaski - end of 2017 now - and rebar (2) has been [deprecated](https://github.com/rebar/rebar/commit/a4d835bdea621f6d2ebc463f773fec620ee12581) since Apr 14, 2016.

michalwski:
I'm closing this one since the conversation is dead and similar changes have been done in #62

PR #60, opened 2018-03-09
Update escalus to newest version

bszaf:


relx.config
```diff
@@ -8,7 +8,6 @@
   {exml, load},
```
mentels: Why do we get rid of these dependencies?

mentels:
And the Travis failed..

relx.config
bszaf: Release will not be build otherwise. It seems that deps of deps were added into `relx.config`

michalwski:
This was superseded by #60

PR #62, opened 2018-04-05
Switch to rebar3 and get compatible to erlang 20

GalaxyGorilla:
Actually this PR should just have been about erlang 20 but for some reasons (probably local env problems) I couldn't handle the old relx binary. So I thought I just update to rebar3 too and spare me the bug searching.

I hope I did not break anything for esl's toolchains. Can you check?


GalaxyGorilla:
I have to do some testing too today, so I set this as WiP.

GalaxyGorilla:
To get dialyzer running I set some ignore flags. Also dependencies are not dialyzed for now since gun has an issue which should be fixed upstream.

GalaxyGorilla:
As a sidenote: I don't know why dialyzer was integrated with CT here but this is the first thing I would revert in the future. There are already dialyzer targets in the Makefile.

michalwski:
Hi @GalaxyGorilla, thanks for the effort! We really appreciate your work.

>  I don't know why dialyzer was integrated with CT here but this is the first thing I would revert in the future. There are already dialyzer targets in the Makefile.

Dialyzer was integrated with CT in order to fail the travis job if there are any errors emitted. We did this to make sure that we keep type and function spec up to date.

rebar.config
```diff
@@ -1,25 +1,39 @@
     {erl_opts, [debug_info,
-                warn_missing_spec,
+                %warn_missing_spec,
```
michalwski: What is the root cause behind disabling this flag?

GalaxyGorilla: I guess I disabled this to prevent a wall of warnings every time i executed something. I will put it back in place.

GalaxyGorilla:
@michalwski Is that really something that is not possible with Makefile targets? I have no experience with travis-ci ... but gitlab-ci, go cd etc. have absolutely no problem with this. It should just care about return codes not equal to 0.

michalwski:
It is possible with travis as well. Currently there is no need to put dialyzer inside CT. 

If you want to remove dialyzer form CT but still have it fail the job you can add `make dialyzer` or `./rebar3 dialyzer` to `scripts` in `.travis.yml`. Like in the following example:

```yml
script:
    - make test
   - ./rebar3 dialyzer
```

GalaxyGorilla:
@michalwski Can we make that in another PR? I can also do this immediately after you merge this (if you are happy about the rest). This PR ist just intended to get amoc ready for erlang 20.

michalwski:
@GalaxyGorilla sure, this can done in a separate PR

GalaxyGorilla:
@michalwski here you go :)

michalwski:
Thanks, it looks good to me. I'd like to wait for my colleagues using Amoc in our continuous load testing tool to see the chenges and comment if this will break existing workflows.

PR #63, opened 2018-04-09
Remove dialyzer from CT

GalaxyGorilla:
As promised in #62 ;)

I did some formatting/refactoring of the Makefile, I hope that's fine.


.travis.yml
```diff
@@ -13,7 +13,8 @@ addons:
             - libexpat1-dev
 
 install: make compile
-script: make test
+script:
+    - make test dialyzer
```
michalwski: I think it be better to run test and dialyzed separately. The log in Travis will be a bit easier to analyse IMO.

GalaxyGorilla:
@michalwski fixed!

PR #66, opened 2018-05-10
Change Docker image building process

bszaf:
Motivation
---
- It was unable to have automated Docker image build for different brach than master with **one** Dockerfile.
- Amoc source was cloned which building Docker image instead of copying existing one.

Explanation
--- 
Previously Amoc source code was retrieved in Dockerfile with cloning code from github. It did not allow to trigger automated builds for other branch than master, because it was always cloning master.

To avoid it, we can copy code into build image. To make it possible it is necessary to move Dockerfile into project root directory. It allows to use `COPY . /amoc_build`.

There is also new file .dockerignore which is similiar to .gitignore. It allows to specify files which should not be copied into build context while building docker image.

When this is merged `Dockerfile` location have to be changed at `hub.docker.com`.


README.md
```diff
@@ -207,6 +207,45 @@ There is available REST API for AMOC where we can:
 * check status of nodes in AMOC cluster
 
 API is described [here](REST_API_DOCS.md). You can also get an access to REST API by running AMOC and go to `(address:port)/api-docs`.
-### Docker
 
-TODO
+## Docker
+
+Building local Docker image may be started with:
```
arkgil: `To build a Docker image with Amoc, run the following command from the root of the repository:` - what do you think about it?

README.md
```diff
@@ -207,6 +207,45 @@ There is available REST API for AMOC where we can:
...
+```
+docker build -f docker/Dockerfile -t tag .
+```
+It is important to start building at project root (it is indicated with dot `.`
```
arkgil: I don't think these two paragraphs are necessary - they describe capabilities of Docker which we're using in the case of Amoc, but aren't specific to it. IMO such information shouldn't be kept in the README.

README.md
```diff
@@ -207,6 +207,45 @@ There is available REST API for AMOC where we can:
...
 
 API is described [here](REST_API_DOCS.md). You can also get an access to REST API by running AMOC and go to `(address:port)/api-docs`.
-### Docker
 
-TODO
+## Docker
+
+Building local Docker image may be started with:
+```
+docker build -f docker/Dockerfile -t tag .
```
arkgil: `-f Dockerfile`: Dockerfile is now placed at the root of the repo.

arkgil: Oh, and maybe `-t image`, to be consistent with the commands below?

README.md
```diff
@@ -207,6 +207,45 @@ There is available REST API for AMOC where we can:
...
+
+Dockerfile leverages multi-stage build feature (more reading [here](https://docs.docker.com/develop/develop-images/multistage-build/)).
+It allows to define intermediate container inside of Dockerfile, which will
+be used just for installing compilation time dependencies and compiling. When
+release is ready inside of temporary container, it is possible to copy just
+the release into fresh image. It will become final image.
+
+When image is ready it may be started just with
+```
+docker run -d image
```
arkgil: I get the following output when running the container:

```
+ SYSCONFIG=/home/amoc/amoc/releases/0.9.0/sys.config
+ sed -e s/AMOC_GRAPHITE_HOST// /sys.config.template
+ sed -i -e s/AMOC_GRAPHITE_PORT// /home/amoc/amoc/releases/0.9.0/sys.config
+ sed -i -e s/AMOC_PREFIX// /home/amoc/amoc/releases/0.9.0/sys.config
+ sed -i -e s/AMOC_HOSTS// /home/amoc/amoc/releases/0.9.0/sys.config
+ /sbin/my_init
*** Running /etc/my_init.d/00_regen_ssh_host_keys.sh...
*** Running /etc/rc.local...
*** Booting runit daemon...
*** Runit started as PID 14
mkdir: cannot create directory ‘/home/amoc/amoc/log’: Permission denied
mkdir: cannot create directory ‘/home/amoc/amoc/log’: Permission denied
mkdir: cannot create directory ‘/home/amoc/amoc/log’: Permission denied
mkdir: cannot create directory ‘/home/amoc/amoc/log’: Permission denied
```

.dockerignore
```diff
@@ -0,0 +1,20 @@
+*.css
+*.html
+*.js
```
arkgil: I don't understand why we need to ignore those, but better keep the Web stuff out of our lawn 😄

bszaf: I simply copied .gitignore and added a few filenames :P

README.md
bszaf: I think we may keep first paragraph, because it describes what are requirements for building Amoc image. The second one could be removed.

arkgil: The only requirement is to be at the project root, which can be described with a single sentence. I just think that the README should be focused on usage, and not internals. Anyway, I'm fine with keeping the first paragraph.

bszaf:
@arkgil Please take a look at it again.

Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+        make \
+        gcc \
+        clang \
+        libexpat1-dev \
+        wget \
+        iproute2 && \
+    wget https://packages.erlang-solutions.com/erlang-solutions_1.0_all.deb && \
+    dpkg -i erlang-solutions_1.0_all.deb && \
+    apt-get update && \
+    apt-get install -y esl-erlang=1:18.3.4
```
michalwski: When do we upgrade do 19.3?

Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+
+COPY . /amoc_build
+
+RUN cd amoc_build && \
+    make rel && \
+    rm _build/default/rel/amoc/*tar.gz
+
+FROM phusion/baseimage
```
michalwski: Why is it duplicated?

Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+MAINTAINER Erlang Solutions <mongoose-im@erlang-solutions.com>
```
michalwski: duplication?

Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+
+RUN useradd -ms /bin/bash amoc
```
michalwski: duplication?

README.md
```diff
@@ -207,6 +207,46 @@ There is available REST API for AMOC where we can:
...
 API is described [here](REST_API_DOCS.md). You can also get an access to REST API by running AMOC and go to `(address:port)/api-docs`.
-### Docker
 
-TODO
+## Docker
+
+To build a Docker image with Amoc, run the following command from the root of
+the repository:
+```
+docker build -f Dockerfile -t image .
```
michalwski: I think it'd better to put more meaningful tag in the example `image` can be applied to anything and people tend to just copy and paste commands when try things.

README.md
```diff
@@ -207,6 +207,46 @@ There is available REST API for AMOC where we can:
...
+```
+It is important to start building at project root (it is indicated with dot `.`
+at the end of command). It will set build context at the project root. Dockerfile
+commands expects a context to be set like that:
+ - it copies **current** source code into container to compile it.
+ - It looks for files in `docker/` relative path.
+
+When image is ready it may be started just with
+```
+docker run -d image
```
michalwski: another `image` tag name which could be changed to sth indicating this is amoc's image

README.md
```diff
@@ -207,6 +207,46 @@ There is available REST API for AMOC where we can:
...
+```
+
+However, you may want to use Amoc HTTP API for uploading and starting scenarios.
+In this case port 4000 should be published.
+```
+docker run -d -p 4000:4000 image
```
michalwski: image image

README.md
```diff
@@ -207,6 +207,46 @@ There is available REST API for AMOC where we can:
...
+```
+
+Amoc logs are connected with container standard output, so it is possible
+to get Amoc logs from running container with:
+```
+docker logs amoc_container
```
michalwski: This doesn't look consistent with other examples. Other examples didn't set container's name to `amoc_container`

README.md
```diff
@@ -207,6 +207,46 @@ There is available REST API for AMOC where we can:
...
+```
+
+Amoc is able to report metrics to Graphite. It is possible to set up
+reporting endpoint by passing following envromental variables:
+```
+docker run -e AMOC_GRAPHITE_HOST=192.168.0.1 -e AMOC_GRAPHITE_PORT=2003 -d image
```
michalwski: image image

docker/run.sh
```diff
@@ -2,6 +2,9 @@
 
 set -x
 
+AMOC_GRAPHITE_HOST=${AMOC_GRAPHITE_HOST:-127.0.0.1}
```
michalwski: I'm not sure that default endpoint for graphite host should be 127.0.0.1:2003. Is it started in the container?

Dockerfile
bszaf: Regarding duplication comments: It is new Docker feature https://docs.docker.com/develop/develop-images/multistage-build/. It allows to create "intermediate" container, which is used just for installing compile time dependencies and compilation. When release is ready inside of temporary container, it is possible to copy just the release into fresh image. It will become final image.

docker/run.sh
bszaf: We need to provide any address. Otherwise Amoc inside container would crash. We have fixed `docker/config/sys.config.template`. If we do not provide `AMOC_GRAPHITE_PORT`, it would have invalid syntax. 

I just added `AMOC_GRAPHITE_HOST` to make it more consistent.

When wrong address it provided (like default one), just `exometer` application will crash. Amoc is still operable.

Dockerfile
bszaf: It is not scope of this PR. 
We should test it with 19.3 and create another PR.

michalwski:
@bszaf all clear now, thanks!

PR #70, opened 2019-01-08
Add pubsub simple scenario

aleklisi:
This PR introduces the following changes:

- Fix bug (`escalus_connection:start` was expected a tuple with 4 fields instead of 3)

- Simple scenario for `pubsub`

### DIFF
--- a/scenarios/pubsub_simple.erl
+++ b/scenarios/pubsub_simple.erl
@@ -12,20 +12,6 @@
 -define(DELAY_BETWEEN_MESSAGES, 100).
 -define(WAIT_FOR_NODES, 10000).
 
--define(NODE_CREATE_TIME, [amoc, times, node_create_time]).
--define(SUBSCRIBE_TIME, [amoc, times, subscribe_time]).
--define(PUBLISH_TIME, [amoc, times, publish_time]).
-
--define(NODE_CREATE_FAILS_COUTER, [amoc, counters, node_create_failed]).
--define(SUBSCRIBE_FAILS_COUTER, [amoc, counters, subscribe_failed]).
--define(PUBLISH_FAILS_COUTER, [amoc, counters, publish_failed]).
-
--define(NODE_CREATE_SUCCESS_COUTER, [amoc, counters, node_create_success]).
--define(SUBSCRIBE_SUCCESS_COUTER, [amoc, counters, subscribe_success]).
--define(PUBLISH_SUCCESS_COUTER, [amoc, counters, publish_success]).
--define(MESSAGE_TTD_CT, [amoc, times, message_ttd]).
-
-
 -export([init/0]).
 -export([start/1]).
 
@@ -55,28 +41,13 @@ start(Id) ->
 %% --- Metrics ---------------------------------------------------------------------------------------------
 
 init_metrics() ->
-    Stats =
-    [
-        {?NODE_CREATE_FAILS_COUTER, spiral},
-        {?SUBSCRIBE_FAILS_COUTER, spiral},
-        {?PUBLISH_FAILS_COUTER, spiral},
-        {?NODE_CREATE_SUCCESS_COUTER, spiral},
-        {?SUBSCRIBE_SUCCESS_COUTER, spiral},
-        {?PUBLISH_SUCCESS_COUTER, spiral},
-        {?NODE_CREATE_TIME, histogram},
-        {?SUBSCRIBE_TIME, histogram},
-        {?PUBLISH_TIME, histogram},
-        {?MESSAGE_TTD_CT, histogram}
-    ],
-    [create_stat(Path, GraphType) ||  {Path, GraphType} <- Stats].
-
-create_stat(Path, spiral) ->
-    exometer:new(Path, spiral),
-    exometer_report:subscribe(exometer_report_graphite, Path, [one, count], 1000);
-
-create_stat(Path, histogram) ->
-    exometer:new(Path, histogram),
-    exometer_report:subscribe(exometer_report_graphite, Path, [mean, min, max, median, 95, 99, 999], 1000).
+    Counters = [ node_creation_success, node_creation_failure, subscription_failure,
+                 publication_failure, subscription_success, publication_success ],
+    Times = [ node_creation, subscription, publication, message_ttd ],
+    lists:foreach(fun(Metric) ->
+                          amoc_metrics:init(counters, Metric) end, Counters),
+    lists:foreach(fun(Metric) ->
+                          amoc_metrics:init(times, Metric) end, Times).
 %% ----------------------------------------------------------------------------------------------------------
 
 assign_role(Id, CreatorsNumber) when Id < CreatorsNumber ->
@@ -143,7 +114,7 @@ subscriber_loop(Client) ->
             TimeStampBin = exml_query:attr(Item, <<"id">>),
             TimeStamp = binary_to_integer(TimeStampBin),
             TTD =  os:system_time(microsecond) - TimeStamp,
-            exometer:update(?MESSAGE_TTD_CT, TTD),
+            amoc_metrics:update_time(message_ttd, TTD),
             lager:info("~n~n~p~n~n", [Stanza])
     catch
         Error:Reason ->
@@ -236,10 +207,10 @@ create_pubsub_node(Client) ->
 
     case escalus_pred:is_iq_result(CreateNodeResult) of
         true ->
-            exometer:update(?NODE_CREATE_SUCCESS_COUTER, 1),
-            exometer:update(?NODE_CREATE_TIME, CreateNodeTime);
+            amoc_metrics:update_counter(node_create_success, 1),
+            amoc_metrics:update_time(node_create_time, CreateNodeTime);
         Error ->
-            exometer:update(?NODE_CREATE_FAILS_COUTER, 1),
+            amoc_metrics:update_counter(node_create_fails, 1),
             lager:error("Error creating node: ~p", [Error]),
             exit(connection_failed)
     end,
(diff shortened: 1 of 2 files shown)


aleklisi:
This PR requires adding mod pubsub to MongooseIM config:
    `{mod_pubsub, [
                 {ignore_pep_from_offline, false},
                 {backend, mnesia},
                 {nodetree, <<"dag">>},
                 {last_item_cache, mnesia},
                 {max_items_node, 1000},
                 {plugins, [<<"flat">>, <<"pep">>]
                 }
  ]},`

scenarios/pubsub_simple.erl
```diff
@@ -0,0 +1,182 @@
...
+    pg2:join(?GROUPNAME, self()),
+    Client = connect_amoc_user(1),
+    Nodes = [create_pubsub_node(Client) || _ <- lists:seq(1, ?NUMBER_OF_PUBSUB_NODES)],
+    provide_nodes_names(Client, Nodes);
+
+start(Id) ->
+    pg2:create(?GROUPNAME),
+    Client = connect_amoc_user(Id),
+    Nodes = get_nodes(3, Client),
+    case Id rem 2 of
```
michalwski: It'd be very helpful if we could somehow parametrize number of publishers and subscribers per node.

scenarios/pubsub_simple.erl
```diff
@@ -0,0 +1,182 @@
...
+-export([init/0]).
+-export([start/1]).
+
+-spec init() -> ok.
+init() -> ok.
+
+-spec start(integer()) -> ok.
+
+start(1) ->
+    pg2:create(?GROUPNAME),
```
michalwski: That's neat! 🏆

scenarios/pubsub_simple.erl
```diff
@@ -0,0 +1,182 @@
...
+            lager:info("Client: Starting subscriber\n", []),
+            work_as_subscriber(Id, Client, Nodes)
+    end,
+    ok.
+
+provide_nodes_names(SelfClient, Nodes) ->
+    receive
+        {what_nodes, PID, Client} ->
+            AffChange = [{Client, <<"publisher">>}],
+            [ set_affiliations(SelfClient, Node, AffChange) || Node <- Nodes],
```
michalwski: As far as I remember, the nodes can be configured (while creating) in such a way that publishers can join by themselves, am I right?

scenarios/pubsub_simple.erl
aleklisi: Well, I could parametrize publisher subscriber ratio with no problem :)

priv/app.config
```diff
@@ -1,5 +1,9 @@
 [
- {amoc, [{interarrival, 75}]},
+ {amoc, [{interarrival, 75},
+         {pubsub_nodes, 5},
```
michalwski: I think it's not a good idea to put scenario specific config in the repository. If we need to read it from the app config I'd rather set default values in the scenario when reading them.

scenarios/pubsub_simple.erl
```diff
@@ -0,0 +1,269 @@
...
+
+provide_nodes_names(SelfClient, Nodes) ->
+    receive
+        {what_nodes, PID} ->
+            PID ! {nodes, Nodes}
+    end,
+    provide_nodes_names(SelfClient, Nodes).
+
+get_nodes(0, _) ->
+    loger:error("Client: NO NODES PROVIDED"),
```
michalwski: probably `lager`, right?

scenarios/pubsub_simple.erl
```diff
@@ -0,0 +1,269 @@
...
+            escalus:wait_for_stanza(Client, ?WAIT_FOR_STANZA_TIMEOUT)
+        end),
+
+    case escalus_pred:is_iq_result(CreateNodeResult) of
+        true ->
+            exometer:update(?NODE_CREATE_SUCCESS_COUTER, 1),
+            exometer:update(?NODE_CREATE_TIME, CreateNodeTime);
+        Error ->
+            exometer:update(?NODE_CREATE_FAILS_COUTER, 1),
+            loger:error("Error creating node: ~p", [Error]),
```
michalwski: probably `lager`, right?

scenarios/pubsub_simple.erl
```diff
@@ -0,0 +1,269 @@
...
+        fun() ->
+            escalus:wait_for_stanza(Client, ?WAIT_FOR_STANZA_TIMEOUT)
+        end),
+    case escalus_pred:is_iq_result(SubscribeResult) of
+        true ->
+            exometer:update(?SUBSCRIBE_SUCCESS_COUTER, 1),
+            exometer:update(?SUBSCRIBE_TIME, SubscribeTime);
+        Error ->
+            exometer:update(?SUBSCRIBE_FAILS_COUTER, 1),
+            loger:error("Error subscribing node ~p filed: ~p", [Node, Error]),
```
michalwski: probably `lager`, right?

scenarios/pubsub_simple.erl
```diff
@@ -0,0 +1,269 @@
...
+
+id(Client, {NodeAddr, NodeName}, Suffix) ->
+    UserName = escalus_utils:get_username(Client),
+    list_to_binary(io_lib:format("~s-~s-~s-~s",
+				 [UserName, NodeAddr, NodeName, Suffix])).
+
+user_spec(ProfileId, Password, Res) ->
+    [ {username, ProfileId},
+      {server, ?HOST},
+      {host, pick_server()},
```
michalwski: `pick_server()` in current form (on Tide) returns sth like this:
```erlang
[{host, MONGOOSE_IP_ADDR},
 {port, MONGOOSE_PORT}]
```
So this config will not work on Tide currently.

scenarios/pubsub_simple.erl
```diff
@@ -0,0 +1,269 @@
...
+      {password, Password},
+      {carbons, false},
+      {stream_management, false},
+      {resource, Res}
+    ].
+
+make_user(Id, R) ->
+    BinId = integer_to_binary(Id),
+    ProfileId = <<"user", BinId/binary>>,
```
michalwski: On Tide it's `user_` prefix.

scenarios/pubsub_simple.erl
```diff
@@ -0,0 +1,269 @@
...
+    Password = <<"password", BinId/binary>>,
```
michalwski: On Tide it's `password_` prefix.

scenarios/pubsub_simple.erl
aleklisi: right

scenarios/pubsub_simple.erl
```diff
@@ -138,6 +139,13 @@ connect_amoc_user(Id, Resource) ->
     {ok, Client, _} = escalus_connection:start(Cfg),
     Client.
 
+create_pubsub_nodes(Id) ->
+    pg2:create(?GROUPNAME),
+    pg2:join(?GROUPNAME, self()),
+    Client = connect_amoc_user(1),
```
aleklisi: 1 -> Id ??

aleklisi: good point, thanks ;P

scenarios/pubsub_simple.erl
aleklisi: right

scenarios/pubsub_simple.erl
aleklisi: right

aleklisi:
remove unused macros.

scenarios/pubsub_simple.erl
```diff
@@ -0,0 +1,320 @@
...
+create_stat(Path, spiral) ->
+    exometer:new(Path, spiral),
+    exometer_report:subscribe(exometer_report_graphite, Path, [one, count], 1000);
+
+create_stat(Path, histogram) ->
+    exometer:new(Path, histogram),
+    exometer_report:subscribe(exometer_report_graphite, Path, [mean, min, max, median, 95, 99, 999], 1000).
+%% ----------------------------------------------------------------------------------------------------------
+
+assign_role(Id, CreatorsNumber) when Id < CreatorsNumber ->
```
aleklisi: `Id < CreatorsNumber`  to 
`Id =< CreatorsNumber` ?

scenarios/pubsub_simple.erl
```diff
@@ -0,0 +1,288 @@
...
+        subscriber ->
+            subscriber_start(Client)
+    end.
+
+%% --- Metrics ---------------------------------------------------------------------------------------------
+
+init_metrics() ->
+    Counters = [ node_create_success, node_create_failed, subscribe_failed,
+                 publish_failed, subscribe_success, publish_success ],
+    Times = [ node_create_time, subscribe_time, publish_time, message_ttd ],
```
michalwski: Since all metrics measuring exec time of sth are prefixed with `[amoc, tide]` I think there is no need to add the `_time` suffix. Additionally I suggest to change counter and time metric names a bit like below (this is not complete list, I imagine you can figure out how to name other metrics based on that):
* `node_creation`
* `node_subscription`
* `item_publication`

I'm aware that this will break your current dashboards, on the other hand I'd like to see these names improved sooner than later. You can also refuse to change because of the existing work you already did. Just let me know :)

scenarios/pubsub_simple.erl
```diff
@@ -0,0 +1,288 @@
...
+
+try_get_member([]) ->
+    Members = pg2:get_members(?GROUPNAME),
+    try_get_member(Members);
+try_get_member(Members) ->
+    L = length(Members),
+    N = erlang:phash2(self(), L) + 1,
+    lists:nth(N, Members).
+
+from_now({MegaSecs,Secs,Usecs}) ->
```
michalwski: There is no need for that functionality. You can:
1. Use `os:system_time(microsecond)` to get current timestamp in microsec, or:
2. Use https://github.com/esl/usec which was made exactly for this in times where there was no `os:system_time` API.

scenarios/pubsub_simple.erl
zofpolkowska: We would have to adjust the dashboard to the new naming convention (https://github.com/esl/MongooseIM/pull/2193) anyway

michalwski:
Thanks!

PR #72, opened 2019-01-15
Extract common parts to helper modules

michalwski:
* improved `amoc_metrics` API
* new `amoc_xmpp` module with helper function
* new `amoc_xmpp_handlers` module with common handlers

### DIFF
--- a/scenarios/mongoose_pubsub_docker.erl
+++ b/scenarios/mongoose_pubsub_docker.erl
@@ -20,22 +20,18 @@
 -define(WAIT_FOR_STANZA_TIMEOUT, 10000).
 -define(PUBSUB_ADDR, <<"pubsub.localhost">>).
 
--define(PUBSUB_NODES_CT, [amoc, counters, pubsub_nodes]).
--define(SUBSCRIPTIONS_CT, [amoc, counters, subscriptions]).
--define(ITEMS_SENT_CT, [amoc, counters, items_sent]).
--define(ITEMS_RECEIVED_CT, [amoc, counters, items_received]).
+-define(PUBSUB_NODES_CT, pubsub_nodes).
+-define(SUBSCRIPTIONS_CT, subscriptions).
+-define(ITEMS_SENT_CT, [counters, items_sent]).
+-define(ITEMS_RECEIVED_CT, [counters, items_received]).
 
 -spec init() -> ok.
 init() ->
     set_env_interarrival(),
-    exometer:new(?PUBSUB_NODES_CT, counter),
-    exometer_report:subscribe(exometer_report_graphite, ?PUBSUB_NODES_CT, [value], 10000),
-    exometer:new(?SUBSCRIPTIONS_CT, counter),
-    exometer_report:subscribe(exometer_report_graphite, ?SUBSCRIPTIONS_CT, [value], 10000),
-    exometer:new(?ITEMS_SENT_CT, spiral),
-    exometer_report:subscribe(exometer_report_graphite, ?ITEMS_SENT_CT, [one, count], 10000),
-    exometer:new(?ITEMS_RECEIVED_CT, spiral),
-    exometer_report:subscribe(exometer_report_graphite, ?ITEMS_RECEIVED_CT, [one, count], 10000),
+    amoc_metrics:new_counter(?PUBSUB_NODES_CT),
+    amoc_metrics:new_counter(?SUBSCRIPTIONS_CT),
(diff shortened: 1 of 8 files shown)


scenarios/mongoose_pubsub_docker.erl
```diff
@@ -20,22 +20,18 @@
...
     set_env_interarrival(),
-    exometer:new(?PUBSUB_NODES_CT, counter),
-    exometer_report:subscribe(exometer_report_graphite, ?PUBSUB_NODES_CT, [value], 10000),
-    exometer:new(?SUBSCRIPTIONS_CT, counter),
-    exometer_report:subscribe(exometer_report_graphite, ?SUBSCRIPTIONS_CT, [value], 10000),
-    exometer:new(?ITEMS_SENT_CT, spiral),
-    exometer_report:subscribe(exometer_report_graphite, ?ITEMS_SENT_CT, [one, count], 10000),
-    exometer:new(?ITEMS_RECEIVED_CT, spiral),
-    exometer_report:subscribe(exometer_report_graphite, ?ITEMS_RECEIVED_CT, [one, count], 10000),
+    amoc_metrics:new_counter(?PUBSUB_NODES_CT),
```
chrzaszcz: Some names are lists, some are atoms, confusing.

src/amoc_metrics.erl
```diff
@@ -1,66 +1,96 @@
 -module(amoc_metrics).
 
+-export([default_reporting_opts/0]).
 -export([new_histogram/1, new_counter/1, new_spiral/1]).
 -export([new_histogram/2, new_counter/2, new_spiral/2]).
 -export([update_hist/2, update_counter/2, update_spiral/2]).
+-export([messages_spiral_name/0, message_ttd_histogram_name/0]).
 
--type name() :: atom().
+-type name() :: atom() | [atom()].
```
chrzaszcz: Do we need both options here? It seems error-prone.

src/amoc_metrics.erl
```diff
@@ -1,66 +1,96 @@
...
 
--spec reporter_and_interval([proplists:property()]) ->
-    {module(), non_neg_integer()}.
-reporter_and_interval(Opts) ->
-    Reporter = proplists:get_value(reporter, Opts, exometer_report_graphite),
-    Interval = proplists:get_value(interval, Opts, 5000),
-    {Reporter, Interval}.
+-spec messages_spiral_name() -> name().
+messages_spiral_name() ->
+    [counters, messages_sent].
```
chrzaszcz: The `counters` prefix suggests it's a counter but it's a spiral. I think it's confusing.

One of the possible solutions would be prefixing all names with their type and enforcing the pattern `[amoc, Type, ...]` where `Type` is `histogram`, `spiral` or `counter` (I don't think the plural makes much sense here). The open question is whether `...` should be one atom or a list of atoms. We could also get rid of the type prefix completely.

I think the current situation is inconsistent as some prefixes are metric types (`histograms`, `spirals`, `counters`) and some are not (`times`).

src/amoc_metrics.erl
```diff
@@ -1,66 +1,96 @@
...
+name_by_type(histogram) -> histograms;
+name_by_type(spiral) -> spirals;
+name_by_type(counter) -> counters.
+
+
+create_metric_and_maybe_subscribe(ExName, Type, Reporter, Interval, Datapoints) ->
+    ok = exometer:re_register(ExName, Type, []),
+    case lists:keyfind(Reporter, 1, exometer_report:list_reporters()) of
+        {Reporter, _} ->
+            exometer_report:unsubscribe(Reporter, ExName, Datapoints),
```
chrzaszcz: I am not sure if we need to unsubscribe here. Will this fail when already subscribed?

michalwski:
I completely agree with you regarding metric names. The reason I did it this way is to be consistent with current dashboards on Tide. What we can do is to allow passing list as metric name (to be compatible with existing dashboards on Tide) but use consistent metric names (as you suggested) in the example scenarios in this repo.

src/amoc_metrics.erl
michalwski: When already subscribed and we subscribe again we may end up with more than one subscription sending the same metric on different times.

src/amoc_metrics.erl
michalwski: Yes we do, because of the existing dashboards on Tide.

michalwski:
@chrzaszcz I made changes based on your comments.

PR #73, opened 2019-01-16
XMPP common parts

michalwski:
This PR extracts some common parts from scenario files to amoc modules.


scenarios/mongoose_docker_one_to_one.erl
```diff
@@ -1,5 +1,5 @@
 %%==============================================================================
-%% Copyright 2015 Erlang Solutions Ltd.
+%% Copyright 2019 Erlang Solutions Ltd.
```
aleklisi: Should it be changed?

aleklisi: How about:
`Copyright (c) 2015-2019 Erlang Solutions Ltd.`
or
`Copyright 2015 Erlang Solutions Ltd.`
`Copyright 2019 Erlang Solutions Ltd.`
?
Like in https://smallbusiness.chron.com/year-copyright-statement-mean-62027.html
section `Several Years`

aleklisi: I think List option fits best

scenarios/mongoose_in_band_registration.erl
```diff
@@ -1,5 +1,5 @@
 %%==============================================================================
-%% Copyright 2015 Erlang Solutions Ltd.
+%% Copyright 2019 Erlang Solutions Ltd.
```
aleklisi: The same question as before

scenarios/mongoose_pubsub_docker.erl
```diff
@@ -142,64 +142,19 @@ id(Client, {NodeAddr, NodeName}, Suffix) ->
...
-    N = erlang:phash2(self(), S) + 1,
-    element(N, Servers).
-
--spec env_servers() -> {binary()}.
-env_servers() ->
-    List = re:split(os:getenv("AMOC_XMPP_SERVERS"), "\s",
-                    [{return, binary}, trim]),
-    list_to_tuple(List).
-
 -spec set_env_interarrival() -> ok.
```
aleklisi: What is the rule that you follow when providing/not providing specs?

scenarios/mongoose_docker_one_to_one.erl
```diff
@@ -26,108 +25,71 @@
...
-    << ProfileId/binary, "@", Host/binary >>.
-
--spec pick_server() -> binary().
-pick_server() ->
-    Servers = amoc_config:get(xmpp_servers),
-    S = size(Servers),
-    N = erlang:phash2(self(), S) + 1,
-    element(N, Servers).
+-spec send_and_recv_escalus_handlers() -> [{atom(), any()}].
+send_and_recv_escalus_handlers() ->
```
aleklisi: I might be wrong but `send_and_recv_escalus_handlers` is used 4 times in different tests.
How about extracting this function too?

bszaf:
@michalwski I think connection function should be provided, which allows to retry after failure. You can reuse scenario parts from https://github.com/esl/MongooseIM/blob/feature/clt-rabbit-mod_event_pusher/load_test/amoc_scenarios/xmpp_amqp.erl#L607-L634.

Also the aforementioned scenario contains a few XMPP helpers, which may be included in this PR :)

PR #74, opened 2019-01-17
XMPP common parts

michalwski:
This PR aims to simplify XMPP scenario writing by extracting common parts to either escalus or amoc modules.
This PR required esl/escalus#202


src/amoc_xmpp.erl
```diff
@@ -1,7 +1,23 @@
...
+-export([connect_or_exit/2]).
+-export([pick_server/0]).
 
+%% @doc connects and authenticates a user with given id and additional properties
+%% If the passed proplist is empty, a default user spec a created by
+%% make_user function is passed.
+%% If the proplist is not empty it will be merged with the default props.
+%% When a property is defined both in default and passed props,
+%% the one from passed props is used.
+-spec connect_or_exit(amoc_scenario:user_id(), [{atom(), any()}]) ->
```
erszcz:
```suggestion
-spec connect_or_exit(amoc_scenario:user_id(), escalus_users:user_spec()) ->
```

src/amoc_xmpp.erl
```diff
@@ -17,3 +33,48 @@ connect_or_exit(Spec) ->
...
+%% [[{host, "127.0.0.1"}, {port, 5222}], [{host, "127.0.0.1"}, {port, 5223}]]
+%% One of the above proplists is picked and can be added to user's spec.
+%% It's required that the proplists contains at least the `host` property.
+%% Since the `xmpp_servers` config option is read via `amoc_config` API,
+%% it's possible to pass it as an ENV var when starting amoc:
+%% > AMOC_xmpp_servers="[[{host,\"127.0.0.2\"}, {port, 5222}],[{host, \"127.0.0.1\"}, {port, 5223}]]" make console
+%% Default value, if xmpp_servers option is not set, is [{host, "127.0.0.1"}]
+-spec pick_server() -> [proplists:property()].
+pick_server() ->
+    Servers = amoc_config:get(xmpp_servers, [[{host, "127.0.0.1"}]]),
```
erszcz: I think the default should be passed in from the scenario, not assumed by the library, since it's more convenient for scenario readers.

In other words I imagine we ought to use:
- OS vars
- Erlang app vars
- a default from the scenario

instead of a hardcoded default. I.e. have `pick_server/1` instead of `pick_server/0`.

src/amoc_xmpp.erl
```diff
@@ -17,3 +33,48 @@ connect_or_exit(Spec) ->
...
+
+verify(Servers) ->
+    lists:foreach(
+      fun(Proplist) ->
+              true = proplists:is_defined(host, Proplist)
+      end,
+      Servers
+     ).
+
+make_user(Id, Props) ->
```
erszcz:
```suggestion
-spec make_user(amoc_scenario:user_id(), escalus_users:user_spec()) -> escalus_users:user_spec().
make_user(Id, Props) ->
```

erszcz: The suggestion diff didn't quite get that it's just a new spec line added :/

scenarios/simple_rest_api.erl
```diff
@@ -12,85 +12,45 @@ init() ->
...
-      {received_stanza_handlers, [fun log_message/2]}
-    ].
-
 log_message(_Client, Stanza) ->
     lager:warning("~p", [Stanza]),
     true.
 
 send_message(AuthHeader, Id) ->
-    Msg = #{to => full_jid(Id),
+    Msg = #{to => ampc_xmpp_stanzas:make_jid(Id),
```
erszcz:
```suggestion
    Msg = #{to => amoc_xmpp_users:make_jid(Id),
```

src/amoc_xmpp.erl
```diff
@@ -17,3 +33,48 @@ connect_or_exit(Spec) ->
...
+    BinId = integer_to_binary(Id),
+    ProfileId = <<"user_", BinId/binary>>,
```
erszcz:
```suggestion
    ProfileId = amoc_xmpp_users:username(Id),
```

src/amoc_xmpp.erl
michalwski: Yes, it makes sense, I'll change `pick_server` to take the default value as an argument.
Currently the `amoc_config:get/2` works like you described, it first tries to read from the OS var, if it's not set it tries to read from the app vars and then returns default if the option was not in app vars.

rebar.config
```diff
@@ -11,7 +11,7 @@
             {jiffy,             "0.15.2"},
             {trails,            "2.1.0"},
             {cowboy_swagger,    "2.1.0"},
-            {escalus,           {git, "git://github.com/esl/escalus.git", {ref, "f20bee4"}}},
+            {escalus,           {git, "git://github.com/esl/escalus.git", {branch, "amoc-scenarios-common-parts"}}},
```
erszcz: We should update the branch now once the Escalus PR is merged.

michalwski: It's done.

erszcz:
Awesome. I was waiting for Travis to finish.

In general, we're starting to accumulate logic relevant to JIDs in three different projects, so it might be time we extracted that into a single place. It's out of scope of this PR, though.

PR #79, opened 2019-03-05
Allow to terminate a scenario when invalid

mkacper:
It's now possible to terminate a scenario at any moment via two
new scenario callbacks `valid/0` and `terminate/1`. The former
checks if the current scenario is still valid e.g. some metrics are
below a certain treshold and must return `continue` or `{stop, Reason}`.
The latter one implements the actual termination. It's called only
if `valid/0` returned `{stop, Reason}`. The reason is passed to
the terminate callback. When `continue` is returned nothing happens.

The both callbacks are optional. Validation is scheduled only if
both of them are implemented. By default, validation interval
equals 60s. It can be changed by setting `scenario_validation_interval`
application environment.

Validation logic happens in `amoc_controller` process only on a
"master" Amoc node and only in "distributed mode" (scenario has
to be started via `amoc_dist:do/3/4`).

Closes MIM-269.

### DIFF
--- a/src/amoc_controller.erl
+++ b/src/amoc_controller.erl
@@ -7,6 +7,7 @@
 
 -define(SERVER, ?MODULE).
 -define(INTERARRIVAL_DEFAULT, 50).
+-define(CHECKING_INTERVAL_DEFAULT, 60*1000).
 -define(TABLE, amoc_users).
 
 -record(state, {scenario :: amoc:scenario() | undefined,
@@ -38,7 +39,8 @@
          remove/2,
          remove/3,
          users/0,
-         test_status/1]).
+         test_status/1,
+         start_scenario_checking/1]).
 %% ------------------------------------------------------------------
 %% gen_server Function Exports
 %% ------------------------------------------------------------------
@@ -97,6 +99,10 @@ users() ->
 test_status(ScenarioName) ->
     gen_server:call(?SERVER, {status, ScenarioName}).
 
+-spec start_scenario_checking(amoc:scenario()) -> ok | skip | {error, term()}.
+start_scenario_checking(Scenario) ->
+    gen_server:call(?SERVER, {start_scenario_checking, Scenario}).
+
 %% ------------------------------------------------------------------
 %% gen_server Function Definitions
 %% ------------------------------------------------------------------
@@ -116,6 +122,9 @@ handle_call(users, _From, State) ->
     Reply = [{count, ets:info(?TABLE, size)},
              {last, ets:last(?TABLE)}],
     {reply, {ok, Reply}, State};
+handle_call({start_scenario_checking, Scenario}, _From, State) ->
+    Reply = maybe_start_scenario_checking(Scenario),
+    {reply, Reply, State};
 handle_call({status, Scenario}, _From, State) ->
     Res = case does_scenario_exist(Scenario) of
         true -> check_test(Scenario, State#state.scenario);
@@ -138,6 +147,9 @@ handle_cast(_Msg, State) ->
 -spec handle_info(any(), state()) -> {noreply, state()}.
 handle_info({start_scenario, Scenario, UserIds, ScenarioState}, State) ->
     start_scenario(Scenario, UserIds, ScenarioState),
+    {noreply, State};
+handle_info({check, Scenario, Interval}, State) ->
+    handle_check(Scenario, Interval),
     {noreply, State};
 handle_info(_Info, State) ->
     {noreply, State}.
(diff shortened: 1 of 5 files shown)


bartekgorny:
> I have one doubt about this implementation. Currently, we are calling `terminate` callback in the context of `amoc_controller` and it is called only once. Therefore author of scenario needs to have knowledge about amoc internals to actually reach `amoc_user` processes.

Do we really care, in this case, if test is terminated nicely or not? I don't know about other possible use cases, but here if the test is not "valid" anymore then we are done, all we need to do is send a notification somewhere and then we can kill'em'all, it is all going to /dev/null anyway.

mkacper:
@bartekgorny exactly! In our case we don't care about user processes termination. We only need to know that the test is no loner valid and then we wipe all the test infrastructure. In `terminate/1` callback we're going to send an HTTP request to other component that actually wipe the whole load test.

bszaf:
> all we need to do is send a notification somewhere and then we can kill'em'all
> ~ bartekgorny

Okay, if it is about notifying external component, the solution is, nomen omen, valid.

mkacper:
Maybe `continue/0`?

src/amoc_controller.erl
```diff
@@ -231,6 +254,38 @@ init_scenario(Scenario) ->
...
+-spec maybe_start_scenario_validation(amoc:scenario()) -> ok | skip.
+maybe_start_scenario_validation(Scenario) ->
+    Callbacks = [{valid, 0}, {terminate, 1}],
+    case are_callbacks_exported(Scenario, Callbacks) of
+        true ->
+            schedule_scenario_validation(Scenario, validation_interval()),
+            lager:info("validation has been started", []),
+            ok;
+        false ->
+            lager:info("validation is not started as callbacks=~p are not"
```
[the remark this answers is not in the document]
mkacper: I think `info` is fine as the callbacks are optional.

bartekgorny:
Ugh, I don't know. `continue/0` sounds like a callback we execute if we decide to go on (as opposed to `terminate`). Something llike `check_test_state`, maybe? I'm not feeling too creative today.

mkacper:
@bartekgorny I think `continue/0` could be a good choice as we decide if we should continue or stop. Perfect name would be `continue?/0`. I have already changed the code to use words "continue" and "check". Please take a look and let me know what you think.

mkacper:
@bartekgorny shall we merge this? If yes please let me rebase the branch (amend the first comment to align it to the new naming and squash the second one) if not please give me your thoughts on that. Thanks!

src/amoc_controller.erl
```diff
@@ -208,6 +220,17 @@ handle_do(Scenario, UserIds, State) ->
...
     end.
 
+-spec handle_check(amoc:scenario(), non_neg_integer()) -> any().
+handle_check(Scenario, Interval) ->
+    case Scenario:continue() of
+        continue ->
+            schedule_scenario_checking(Scenario, Interval),
+            continue;
+        {stop, Reason} ->
+            lager:info("scenario is no longer valid, reason=~p", [Reason]),
```
bartekgorny: scenario should be terminated

src/amoc_controller.erl
```diff
@@ -231,6 +254,38 @@ init_scenario(Scenario) ->
...
+        Error ->
+            lager:error("scenario module ~p cannot be found, reason: ~p",
+                        [Scenario, Error]),
+            false
+    end.
+
+-spec schedule_scenario_checking(amoc:scenario(), non_neg_integer()) ->
+    reference().
+schedule_scenario_checking(Scenario, Interval) ->
+    erlang:send_after(Interval, ?SERVER, {check, Scenario, Interval}).
```
bartekgorny: I don't trust loops done this way, I was recently bitten in the backside by such code. Do we know what is going to happen if `Scenario:continue()` fails? Will the whole test crash? Will it terminate gracefully? Or will it keep running forever, without any more checking?

src/amoc_scenario.erl
```diff
@@ -4,9 +4,13 @@
...
 
+-optional_callbacks([valid/0, terminate/1]).
+
 -export_type([user_id/0]).
 
 -type user_id() :: non_neg_integer().
 
 -callback init() -> ok | {error, Reason :: term()}.
 -callback start(user_id()) -> any().
+-callback valid() -> continue | {stop, Reason :: term()}.
```
bartekgorny: so is the callback name `valid` or `continue`?

src/amoc_controller.erl
```diff
@@ -288,6 +343,10 @@ node_userids(Start, End, Nodes, NodeId) when is_integer(Nodes), Nodes > 0,
 interarrival() ->
     amoc_config:get(interarrival, ?INTERARRIVAL_DEFAULT).
 
+-spec validation_interval() -> integer().
+validation_interval() ->
+    amoc_config:get(scenario_checking_interval, ?CHECKING_INTERVAL_DEFAULT).
+
```
bartekgorny: More consistent naming would be better - `validation_interval/0` returns config param `scenario_checking_interval`.

src/amoc_controller.erl
mkacper: >I don't trust loops done this way, I was recently bitten in the backside by such code.
@bartekgorny what do you mean here?

I'm guessing that `amoc_controller` would crash and be restarted by the supervisor and test would last forever as checking wouldn't work after restart. I did not catch exceptions when calling the callbacks as I wanted to stick to the rest of the code. What do you suggests then?

PR #87, opened 2019-08-29
Improve config helper module

michalwski:
This PR:
* merges the `config` module in helper with `amoc_config`
* removes the functionality to store scenario parameters in process dictionary, instead they can be passed as a variable
* adds tests


src/amoc_config.erl
```diff
@@ -5,41 +5,103 @@
...
     case os:getenv("AMOC_" ++ erlang:atom_to_list(Name)) of
         false ->
             application:get_env(amoc, Name, Default);
         Value ->
             parse_value(Value)
-    end.       
+    end.
 
--spec fetch(atom()) -> any().
-fetch(Name) ->
```
DenysGonchar: are you sure that none of the scenarios is using it?

michalwski: I checked, and non in the repo uses this function. If there is other scenario outside the repo, I don't know.

src/amoc_config.erl
```diff
@@ -5,41 +5,103 @@
...
+                       {true, false} ->
+                           lager:error("Invalid value for ~p", [Name]),
+                           {false, bad_value};
+                       {false, false} ->
+                           lager:error("Invalid default value & value for ~p", [Name]),
+                           {false, bad_value_bad_default_value}
+                   end,
+    {Name, Value, Verification}.
+
+verify_value(_, none) -> true;
```
DenysGonchar: lets create amoc_config_validation module with all these:
* none/1
* positive_integer/1
* nonnegative_integer/1
* bitstring/1

and have a general implementation like this one:
```

verify_value(Value, Atom) when is_atom(Atom) ->
    Fun = fun(Value) ->
            erlang:apply(amoc_config_validation, Atom, [Value])
          end,
    call_verify_fun(Fun, Value)
verify_value(Value, [_ | _] = NonemptyList) ->
    is_one_of(Value, NonemptyList);
verify_value(Value, Fun) when is_function(Fun, 1) ->
    call_verify_fun(Fun, Value);
verify_value(_, VerificationMethod) ->
    lager:error("invalid verification method ~p", [VerificationMethod]),
    false.

call_verify_fun(Fun, Value) ->
    try Fun(Value) of
        Bool when is_boolean(Bool) -> Bool;
        _ ->
            lager:error("invalid verification method ~p", [Fun]),
            false
    catch
        C:E:S ->
            lager:error("invalid verification method ~p ~p ~p", [C,E,S]),
            false
    end

```

so it will be easy to extend the list of validation functions. MB even make the module name configurable via erlang application env.

scenarios/pubsub_pep.erl
```diff
@@ -42,10 +42,10 @@
 -spec init() -> {ok, amoc_scenario:state()} | {error, Reason :: term()}.
 init() ->
     init_metrics(),
-    case config:get_scenario_settings(?ALL_PARAMETERS) of
+    case amoc_scenario_config:get_scenario_settings(?ALL_PARAMETERS) of
```
DenysGonchar: please fix scenarios, we don't have `amoc_scenario_config` module anymore

scenarios/pubsub_simple.erl
```diff
@@ -39,7 +39,7 @@
 -spec init() -> {ok, amoc_scenario:state()} | {error, Reason :: term()}.
 init() ->
     init_metrics(),
-    case config:get_scenario_settings(?ALL_PARAMETERS) of
+    case amoc_scenario_config:get_scenario_settings(?ALL_PARAMETERS) of
```
DenysGonchar: please fix scenarios, we don't have amoc_scenario_config module anymore

src/amoc_config.erl
```diff
@@ -5,41 +5,103 @@
...
-         get/2,
-         fetch/1]).
+         get/2]).
+
+-export([parse_scenario_settings/1,
+         get_scenario_parameter/2]).
+
+-export_type([scenario_configuration/0, settings/0]).
+
+-type name() :: atom().
```
DenysGonchar: lets make name be either atom or string

src/amoc_config.erl
```diff
@@ -5,41 +5,103 @@
...
 %% API
 %% ------------------------------------------------------------------
--spec get(atom()) -> any().
+-spec get(name()) -> any().
 get(Name) ->
     get(Name, undefined).
 
--spec get(atom(), any()) -> any().
+-spec get(name(), any()) -> any().
 get(Name, Default) ->
```
DenysGonchar: add support for Name as list value

src/amoc_config.erl
```diff
@@ -5,41 +5,103 @@
...
-         fetch/1]).
+         get/2]).
+
+-export([parse_scenario_settings/1,
+         get_scenario_parameter/2]).
+
+-export_type([scenario_configuration/0, settings/0]).
+
+-type name() :: atom().
+-type env_var() :: atom().
```
DenysGonchar: better make it a `list()`

src/amoc_config.erl
```diff
@@ -5,41 +5,103 @@
...
+-type value() :: any().
+
+-type one_of() :: [any()].
+-type func() :: fun((any()) -> boolean()).
+-type validation() :: func() | one_of() | bitstring | nonnegative_integer | positive_integer | none.
```
DenysGonchar: one_of() | atom() | func()

src/amoc_config.erl
michalwski: Could you give me example where it would be handy?

DenysGonchar: it's just weird to have it as an atom to later on convert it into the list.

src/amoc_config.erl
```diff
@@ -5,41 +5,103 @@
...
+-spec get(name()) -> any().
 get(Name) ->
     get(Name, undefined).
 
--spec get(atom(), any()) -> any().
+-spec get(name(), any()) -> any().
 get(Name, Default) ->
     case os:getenv("AMOC_" ++ erlang:atom_to_list(Name)) of
         false ->
             application:get_env(amoc, Name, Default);
```
DenysGonchar: do we really need this? is it used anyhow anywhere?

src/amoc_config.erl
michalwski: Where is it converted into a list?

src/amoc_config.erl
michalwski: With docker containers it's not used, by in some other cases amoc was configured via application settings. It's better to leave it as is.

src/amoc_config.erl
DenysGonchar: just one line below:
``` case os:getenv("AMOC_" ++ erlang:atom_to_list(Name)) of ```

src/amoc_config.erl
michalwski: I think we are already flexible enough here because of the `fun()` option. If needed another validation module could be passed as a scenario parameter. The validator can specified like:
* `fun erlang:is_bitstring/1` or
* `fun(V) -> erlang:apply(amoc_config:get_scenario_param(validation_mod), validatation_func, [Value]) end`

I know it may not be very convenient but I would wait with such a change until we see how this validation is used in practice. For now we have only two very similar scenarios using it.

PR #88, opened 2019-08-29
Allow to pass extra code paths to the container

michalwski:
The paths with extra code can be passed via `AMOC_EXTRA_CODE_PATHS` env variable.

### DIFF
--- a/docker/run.sh
+++ b/docker/run.sh
@@ -4,12 +4,19 @@ set -x
 
 AMOC_GRAPHITE_HOST=${AMOC_GRAPHITE_HOST:-127.0.0.1}
 AMOC_GRAPHITE_PORT=${AMOC_GRAPHITE_PORT:-2003}
+AMOC_EXTRA_CODE_PATHS=${AMOC_EXTRA_CODE_PATHS:-""}
 
 SYSCONFIG="/home/amoc/amoc/releases/0.9.0/sys.config"
+VMARGS="/home/amoc/amoc/releases/0.9.0/vm.args"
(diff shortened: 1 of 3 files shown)


michalwski:
The docker image is `mongooseim/amoc:docker-with-extra-code-paths`, you can pull and check if you want.

docker/run.sh
```diff
@@ -4,12 +4,19 @@ set -x
...
+VMARGS="/home/amoc/amoc/releases/0.9.0/vm.args"
 sed -e "s/AMOC_GRAPHITE_HOST/${AMOC_GRAPHITE_HOST}/" /sys.config.template > \
     ${SYSCONFIG}
 sed -i -e "s/AMOC_GRAPHITE_PORT/${AMOC_GRAPHITE_PORT}/" ${SYSCONFIG}
 sed -i -e "s/AMOC_PREFIX/${AMOC_PREFIX}/" ${SYSCONFIG}
 sed -i -e "s/AMOC_HOSTS/${AMOC_HOSTS}/" ${SYSCONFIG}
 
+if [[ -n "${AMOC_EXTRA_CODE_PATHS// /}" ]]; then
+    #if the var is not empty and contains sth else then just spaces
+    echo "-pz ${aMOC_EXTRA_CODE_PATHS}" >> ${VMARGS}
```
mkacper: There is a typo here. Should be `AMOC_EXTRA_CODE_PATHS` (the first letter should be upper-case).

mkacper:
I've just checked the image and it works like a charm!

PR #89, opened 2019-08-30
Run xref on CI

michalwski:
This PR adds xref check to CI and fixes a call to undefined function from `amoc_annotations`.


DenysGonchar:
i have added xref in my PR as well, will unmerge it.

PR #92, opened 2019-09-17
Amoc coordinator

DenysGonchar:
amoc_coordinator allows to group users in a simple manner

### DIFF
--- a/src/amoc_coordinator.erl
+++ b/src/amoc_coordinator.erl
@@ -105,13 +105,13 @@ reset(Name) ->
 %%
 %% @end
 %%--------------------------------------------------------------------
--spec(init({timeout, atom(), coordination_timeout_in_sec()} | normalized_coordination_item()) ->
-    {ok, state()}).
+-spec init({timeout, atom(), coordination_timeout_in_sec()} | normalized_coordination_item()) ->
+    {ok, state()}.
 init({timeout, Name, Timeout}) ->
     Pid = spawn(fun() -> timeout_fn(Name, 1000 * Timeout, infinity) end),
     {ok, {timeout, Pid}};
 init(CoordinationItem) ->
-    {ok, Pid} = amoc_coordinator_worker:start(CoordinationItem),
+    {ok, Pid} = amoc_coordinator_worker:start_link(CoordinationItem),
     {ok, {worker, Pid}}.
 
 %%--------------------------------------------------------------------
@@ -123,7 +123,7 @@ init(CoordinationItem) ->
 %%
 %% @end
 %%--------------------------------------------------------------------
--spec(handle_event(Event :: term(), state()) -> {ok, state()}).
+-spec handle_event(Event :: term(), state()) -> {ok, state()}.
 handle_event(Event, {timeout, Pid}) ->
     erlang:send(Pid, Event),
     {ok, {timeout, Pid}};
@@ -147,7 +147,7 @@ handle_event(Event, {worker, Pid}) ->
 %%
 %% @end
 %%--------------------------------------------------------------------
--spec(handle_call(Request :: term(), state()) -> {ok, {error, not_implemented}, state()}).
+-spec handle_call(Request :: term(), state()) -> {ok, {error, not_implemented}, state()}.
 handle_call(_Request, State) ->
     {ok, {error, not_implemented}, State}.
 
@@ -161,7 +161,7 @@ handle_call(_Request, State) ->
 %% @spec terminate(Reason, State) -> void()
 %% @end
 %%--------------------------------------------------------------------
--spec(terminate(any(), state()) -> ok).
+-spec terminate(any(), state()) -> ok.
 terminate(_, {timeout, Pid}) ->
     erlang:send(Pid, terminate), ok;
 terminate(_, {worker, Pid}) ->
--- a/src/amoc_coordinator.erl
+++ b/src/amoc_coordinator.erl
@@ -1,0 +1,193 @@
+-module(amoc_coordinator).
+
+-behaviour(gen_event).
+
+%% API
+-export([start/3, start/2,
+         add/2, add/3,
+         stop/1, reset/1]).
+
+%% gen_event callbacks
+-export([init/1,
+         handle_event/2,
+         handle_call/2,
+         terminate/2]).
+
+-define(DEFAULT_TIMEOUT, 30000). %% 30 seconds
+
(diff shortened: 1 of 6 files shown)


src/amoc_coordinator.erl
```diff
@@ -0,0 +1,206 @@
...
+
+%%--------------------------------------------------------------------
+%% @private
+%% @doc
+%% Whenever a new event handler is added to an event manager,
+%% this function is called to initialize the event handler.
+%%
+%% @end
+%%--------------------------------------------------------------------
+-spec(init({timeout, atom(), coordination_timeout_in_sec()} | normalized_coordination_item()) ->
```
michalwski: What is the benefit of using this `spec` syntax over the other syntax used in previous lines? Is it possible to be consistent?

src/amoc_coordinator.erl
```diff
@@ -0,0 +1,206 @@
...
+%%--------------------------------------------------------------------
+%% @private
+%% @doc
+%% Whenever an event manager receives an event sent using
+%% gen_event:notify/2 or gen_event:sync_notify/2, this function is
+%% called for each installed event handler to handle the event.
+%%
+%% @end
+%%--------------------------------------------------------------------
+-spec(handle_event(Event :: term(), state()) -> {ok, state()}).
```
michalwski: Another spec syntax inconsistency.

src/amoc_coordinator.erl
```diff
@@ -0,0 +1,206 @@
...
+%%--------------------------------------------------------------------
+%% @private
+%% @doc
+%% Whenever an event manager receives a request sent using
+%% gen_event:call/3,4, this function is called for the specified
+%% event handler to handle the request.
+%%
+%% @end
+%%--------------------------------------------------------------------
+-spec(handle_call(Request :: term(), state()) -> {ok, {error, not_implemented}, state()}).
```
michalwski: Another spec syntax inconsistency.

src/amoc_coordinator.erl
```diff
@@ -0,0 +1,206 @@
...
+%% @private
+%% @doc
+%% Whenever an event handler is deleted from an event manager, this
+%% function is called. It should be the opposite of Module:init/1 and
+%% do any necessary cleaning up.
+%%
+%% @spec terminate(Reason, State) -> void()
+%% @end
+%%--------------------------------------------------------------------
+-spec(terminate(any(), state()) -> ok).
```
michalwski: Another spec syntax inconsistency.

src/amoc_coordinator_worker.erl
```diff
@@ -0,0 +1,127 @@
...
+                n = 0 :: non_neg_integer(),
+                actions = [] :: [action()],
+                collect_data = true :: boolean(),
+                accumulator = [] :: [data()]}).
+
+%%%===================================================================
+%%% API
+%%%===================================================================
+
+-spec(start(amoc_coordinator:normalized_coordination_item()) -> {ok, Pid :: pid()}).
```
michalwski: Is it possible to use one spec syntax? At least across single PR.

src/amoc_coordinator_worker.erl
```diff
@@ -0,0 +1,127 @@
...
+start(CoordinationItem) -> gen_server:start(?MODULE, CoordinationItem, []).
```
michalwski: Could you explain why we use `start` and not `start_link` here?

michalwski:
Oh, one more thing. Is this meant to be used in single AMOC node or in a cluster?

src/amoc_coordinator.erl
```diff
@@ -0,0 +1,206 @@
...
+            %%    handle_event(Event, State) for each installed event
+            %%    handler, in the same order as they were added.
+            %%
+            %% in reality the order is reversed, the last added handler
+            %% is executed at first. so to ensure that all the items in
+            %% the plan with NoOfUsers =:= all are executed in the very
+            %% end, we need to add them first.
+            AllItemsHandlers = lists:reverse([Item || {all, _} = Item <- Plan]),
+            [gen_event:add_handler(Name, ?MODULE, Item) || Item <- AllItemsHandlers],
+            [gen_event:add_handler(Name, ?MODULE, Item) || {N, _} = Item <- Plan, is_integer(N)],
```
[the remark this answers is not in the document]
asas12: If I am not mistaken they won't be, as they are denoted with `all` as the first element of the tuple. They will be added from `AllItemsHandlers` and not added in the next line, as they are filtered out based on the first element in the tuple not being an integer.

src/amoc_coordinator_worker.erl
DenysGonchar: it should be started at the Scenario:init(), AFAIR that process is shortlived. And we should use it later from Scenario:start() context.

src/amoc_coordinator.erl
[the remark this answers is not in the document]
michalwski: Right, I didn't notice the guard. Thanks!

DenysGonchar:
> Oh, one more thing. Is this meant to be used in single AMOC node or in a cluster?

in the multi node environment each node can have it's own coordinator, so it will coordinate only the users of the exact slave amoc node

src/amoc_coordinator_worker.erl
michalwski: And do we care if we loose the `amoc_coordinator_worker` process while the test is running?

DenysGonchar: sorry, i got i wrong. yes, this one can and should be changed to `start_link`. it's important to have `gen_event` to be just started, but not linked.

DenysGonchar:
> I think this functionality has some potential, it also has a risk of being not used especially when there is no documentation. I'd like to see some documentation with examples showing benefits of this feature.

i was thinking about creation of some simple example scenario, also it will be used in the (at least) 3 available scenarios. however some text description is good idea.  also it would be nice to describe amoc_throttle as well. so i would prefer to create a separate PR with documentation.

also where would you like me to place such documentation? extend Readme.md? repo's wiki pages? create some docs dir?

michalwski:
> also @michalwski where would you like me to place such documentation? extend Readme.md? repo's wiki pages? create some docs dir?

Right, that's good question. So far we have only Readme.md I think the documentation could be put there for now. Later, we probably need to restructure the doc a move some content to dedicated pages in the `doc` dir.

PR #96, opened 2019-10-07
Adjust codebase to use amoc-xmpp-arsenal - part 2

sebb7:
This PR removes all scenarios, helpers and scenario/xmpp related files from amoc.


rebar.config.script
```diff
@@ -1,21 +1,3 @@
 os:putenv("EXOMETER_PACKAGES", "(basic)"),
```
DenysGonchar: do we need this `rebar.config.script` at all? why can't we remove it?

scenarios/throttle_test.erl
```diff
@@ -1,119 +0,0 @@
-%%%-------------------------------------------------------------------
```
DenysGonchar: this one should stay. it's just a general amoc_throttle example. it doesn't require MIM and can be started on just pure amoc

rebar.config.script
michalwski: I'll check that if we still need it. We needed it some time ago in order to prevent exometer from downloading all possible deps which we didn't need.

src/amoc.app.src
```diff
@@ -8,13 +8,13 @@
...
+        exometer_core,
+        exometer_report_graphite,
+        exometer_report_statsd,
         ssl,
         trails,
         cowboy,
         cowboy_swagger,
-        amqp_client,
         runtime_tools,
         mnesia
```
DenysGonchar: remove this one, as it's needed only for removed iq_metrics helper

PR #97, opened 2019-10-10
Amoc slave rework

zofpolkowska:
This PR includes:
* `amoc_slave` rework (master node is chosen only when the `amoc_dist:do/{3,4}` is executed
* simple scripts to start **AMOC** along with **MongooseIM** in dockerized environment
* the documentation changes will be applied in a separate PR


src/amoc_dist.erl
```diff
@@ -103,7 +103,6 @@ ceil(Number) ->
 ping_node(Node) ->
     case amoc_slave:ping(Node) of
```
michalwski: I think we no longer need the case now when we removed the `amoc_slave:monitor_master` on `pong` result.

zofpolkowska: Yup, we clearly don't need it. I was thinking of removing it, thank for reminding :)

PR #98, opened 2019-10-10
IQ metrics helper

gustawlippa:
Had to change helper to gen_server because amoc doesn't have mnesia. I don't know if it makes any sense.

### DIFF
--- a/scenarios/helpers/iq_metrics.erl
+++ b/scenarios/helpers/iq_metrics.erl
@@ -1,0 +1,108 @@
+-module(iq_metrics).
+
+-export([start/1, request/2, request/3, timeout/2, response/2, new_id/1]).
+
+-record(request, {
+    key :: {Id :: binary(), pid()},
+    name :: name(),
+    time :: integer(),
+    tref :: timer:tref() | undefined | timed_out
+}).
+
+-type name() :: atom().
+
+-define(TABLE, request).
+
+-spec start([name()]) -> ok.
+start(Names) ->
+    {ok, App} = application:get_application(),
+    PrivDir = code:priv_dir(App),
+    application:set_env(mnesia, dir, PrivDir),
+
+    mnesia:create_schema([node()]),
+    application:ensure_started(mnesia),
+    mnesia:create_table(?TABLE, [{attributes, record_info(fields, request)}]),
+
+    Suffixes = [[request], [response], [timeout], [response, result], [response, error], [timeout, result], [timeout, error]],
+    [begin
+         [amoc_metrics:init(counters, [Name | Suffix]) || Suffix <- Suffixes],
+         amoc_metrics:init(times, [Name, response, time])
+     end || Name <- Names],
+    ok.
+
+%% API
+
+-spec request(name(), binary()) -> any().
+request(Name, Id) ->
+    process_request(Name, Id, undefined).
+
+-spec request(name(), binary(), timeout()) -> any().
+request(Name, Id, Timeout) ->
+    {ok, TRef} = timer:apply_after(Timeout, ?MODULE, timeout, [{Id, self()}, no_delete]),
+    process_request(Name, Id, TRef).
+
+-spec timeout({binary(), pid()}, delete | no_delete) -> any();
+             (binary(), delete | no_delete) -> any().
+timeout({_Id, _Pid} = Key, no_delete) ->
+    F = fun() ->
+            case mnesia:read(?TABLE, Key) of
+                [Req] -> mnesia:write(Req#request{tref = timed_out}),
+                         Req;
+                [] -> ok % timed out during success or failure execution
+            end
+        end,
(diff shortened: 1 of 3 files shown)


scenarios/helpers/metrics_helper.erl
```diff
@@ -0,0 +1,121 @@
...
+
+-type name() :: atom().
+
+%% API
+
+-spec start([name()]) -> 'ignore' | {'error',_} | {'ok',pid()}.
+start(Names) ->
+    gen_server:start_link({local, ?SERVER}, ?MODULE, Names, []).
+
+-spec request(name()) -> binary().
```
gustawlippa: this one to remove

scenarios/helpers/metrics_helper.erl
```diff
@@ -0,0 +1,121 @@
...
+%% Internal functions
+
+process_request(Name, Id, TRef) ->
+    Now = os:system_time(microsecond),
+    amoc_metrics:update_counter([Name, request]),
+    Req = #request{key = {Id, self()}, name = Name, time = Now, tref = TRef},
+    ets:insert(?TABLE, Req).
+
+-spec new_id() -> binary().
+new_id() ->
```
gustawlippa: this one to export & add prefix argument

scenarios/helpers/iq_metrics.erl
```diff
@@ -0,0 +1,108 @@
...
+
+-spec start([name()]) -> ok.
+start(Names) ->
+    {ok, App} = application:get_application(),
+    PrivDir = code:priv_dir(App),
+    application:set_env(mnesia, dir, PrivDir),
+
+    mnesia:create_schema([node()]),
+    application:ensure_started(mnesia),
+    mnesia:create_table(?TABLE, [{attributes, record_info(fields, request)}]),
```
michalwski: As far as I understand we want mnesia here in order to have the information replicated to all amoc nodes. Am I correct? If so did you check if table created in this way is in fact replicated to all the nodes?

michalwski: Also I think this table can be ram only. Did you check if this configuration gives us this behaviour?

scenarios/helpers/iq_metrics.erl
```diff
@@ -0,0 +1,108 @@
...
+             (binary(), delete | no_delete) -> any().
+timeout({_Id, _Pid} = Key, no_delete) ->
+    F = fun() ->
+            case mnesia:read(?TABLE, Key) of
+                [Req] -> mnesia:write(Req#request{tref = timed_out}),
+                         Req;
+                [] -> ok % timed out during success or failure execution
+            end
+        end,
+    case mnesia:transaction(F) of
```
michalwski: Do we need transaction here? Would `mnesia:dirty_read` be enough here?

scenarios/pubsub_pep.erl
```diff
@@ -60,13 +60,9 @@ start(Id, Settings) ->
...
 init_metrics() ->
-    Counters = [message,
-                %% node counters
-                node_creation_failure, node_creation_success, node_creation_timeout,
-                %% publication counters
-                publication_query, publication_result, publication_error,
-                publication_success, publication_timeout],
-    Times = [node_creation, publication, message_ttd],
+    iq_metrics:start([node_creation, publication]),
+    Counters = [message],
```
michalwski: The name of a metric can be just an atom like this `message`.

scenarios/helpers/iq_metrics.erl
gustawlippa: When ran locally with this configuration it is ram only but perhaps it should be ensured explicitly

scenarios/helpers/iq_metrics.erl
gustawlippa: As I understand it, we need a transaction, as it may happen that an amoc user reports a response (which results in deleting a record from the table) in between reading and writing here.

scenarios/pubsub_pep.erl
gustawlippa: Sorry, I'm not sure what you mean. There are two groups of metrics handled by iq helper: `node_creation` and `publication`, which are both atoms. Do you mean that i.e. there should be `publication` metric instead of `publication.request`?

scenarios/helpers/iq_metrics.erl
gustawlippa: I think writing to the table can be changed to `dirty` instead of transaction.

scenarios/pubsub_pep.erl
michalwski: As far as I can see the `Counters = [message]` var is used in `amoc_metrics:init`. This function accepts just atom as a name of the function. I thought that it may be a bit more readable if we use just atom as a name of the counter metrics.

PR #103, opened 2019-10-23
Rework scenario documentation

gustawlippa:
Added description for start/2 function which is used in some tests.
Moved description of callback functions from unneeded `Advanced scenario features` to file describing scenarios themselves, so all information about them is in one place.


doc/scenario.md
```diff
@@ -32,6 +34,62 @@ start(Id) ->
...
+start(Id, Settings) ->
+    %% connect user using Settings
+    %% fetch user's history
+    %% send some messages
+    %% wait a little bit
+    %% send some messages again
+    ok.
+```
+
+For more advanced features, `amoc_scenario` behaviour proposes optional callbacks:
```
michalwski: Introducing a header here may make it more readable that a new, important section starts here.

gustawlippa: Is `Optional callbacks` ok? Or something like those `advanced features`?

michalwski: I vote for `advanced features` :)

doc/scenario.md
```diff
@@ -2,15 +2,17 @@
...
+It has to export two callback functions:
+- `init/0` - called only once per test run on every node, at the very beginning.
+It can be used for setting up initial (global) state: metrics, database connections, etc.
+It can return an `ok` or a tuple `{ok, State}` from which the `State` may be passed to every user.
+- `start/1` or `start/2` - describes the actual scenario and is executed for
 each user, in the context of that user's process.
-It takes one argument, which is the given user's unique integer id.
-When the function returns, it is executed again after some delay (60 seconds by default).
+The first argument is the given user's unique integer id.
+The second, which is optional, is the state, as returned from `init/0` function.
```
NelsonVides: > [...] from the `init/0` function.

(article)

doc/scenario.md
```diff
@@ -2,15 +2,17 @@
...
+When `start` function returns, it is executed again after some delay (60 seconds by default).
```
NelsonVides: > When the `start` function returns, [...]

or

> When `start` returns

(The first option is with  a determinate article used to specify which "function", the start function; the second option needs no article because in that one we use the name "start" as a proper name, which doesn't need an article).

doc/scenario.md
```diff
@@ -32,6 +34,62 @@ start(Id) ->
...
+
+start(Id, Settings) ->
+    %% connect user using Settings
+    %% fetch user's history
+    %% send some messages
+    %% wait a little bit
+    %% send some messages again
+    ok.
+```
+### Advanced features
```
NelsonVides: I wouldn't mind an empty line between these two blocks, between the \`\`\` and the ###

doc/scenario.md
```diff
@@ -32,6 +34,62 @@ start(Id) ->
...
+In addition to the required ones, `amoc_scenario` behaviour proposes optional callbacks:
```
NelsonVides: > [...], the `amoc_scenario` behaviour proposes [...]

(article)

doc/scenario.md
```diff
@@ -32,6 +34,62 @@ start(Id) ->
...
+    By default, it is scheduled to run every 60s.
+    This can be changed by setting the `scenario_checking_interval` application environment variable.
+    The checking logic happens in the `amoc_controller` process only on a
+    "master" Amoc node and only in the "distributed mode" (the scenario has
+    to be started via `amoc_dist:do/3/4`).
+
+-   `next_user_batch/2`
+    
+    This callback enables adding users in batches.
+    It is supposed to return strategy for adding users,
```
NelsonVides: > [...] to return a strategy [...]

(article)

doc/scenario.md
```diff
@@ -32,6 +34,62 @@ start(Id) ->
...
+    which is a list of `{Node, NumOfUsers, Interarrival}`.
+    It is then used by `amoc_controller:add_batches/2` function,
```
NelsonVides: > [...] by the `amoc_controller...` [...]

(article)

doc/scenario.md
```diff
@@ -32,6 +34,62 @@ start(Id) ->
...
+    which takes the scenario module and the number of batches as parameters.
+    Parameters passed to the `next_user_batch` callback are
```
NelsonVides: > The parameters passed [...]

(article)

It would be valid to omit the article only if we were talking about parameters in general, like, all parameters to all functions in the world (e.g. "Parameters are given to functions"), but we're talking about the specific parameters of this specific function, so, definite article 😛

doc/scenario.md
```diff
@@ -32,6 +34,62 @@ start(Id) ->
...
+-   `next_user_batch/2`
+    
+    This callback enables adding users in batches.
+    It is supposed to return strategy for adding users,
+    which is a list of `{Node, NumOfUsers, Interarrival}`.
+    It is then used by `amoc_controller:add_batches/2` function,
+    which takes the scenario module and the number of batches as parameters.
+    Parameters passed to the `next_user_batch` callback are 
+    the batch index and the number of users added in the previous batch.
+    Batches are added every batch interval specified by the `add_batch_interval`
```
NelsonVides: > [...] every batch interval as specified [...]

(preposition)

doc/scenario.md
```diff
@@ -32,6 +34,62 @@ start(Id) ->
...
+    application environment variable, which is 5 minutes by default.
+    Adding batches can be scheduled via HTTP API by specifying batches key
```
NelsonVides: I think you meant "batch keys" and not "batches key". I mean, the pluralised word is the leading one — we add keys of type batch. Is that what you meant, right?

PR #104, opened 2019-10-23
adding N of coordinate items as part of coordination event

DenysGonchar:
minor changes to amoc_coordinator


DenysGonchar:
just a bit - event type declaration and regenerate output for the example.
and i notified @aleklisi about these changes in advance, before he finished the doc.

DenysGonchar:
`This means that in other cases handlers are not interested in the newly added N number.`
if you don't need it, it's easy to ignore it. from amoc_coordinator side there is no additional logic to count that number, it's already counted.

PR #105, opened 2019-10-24
Amoc configuratio rework

DenysGonchar:
this PR contains rework of the amoc configuration and simplification of the amoc docker image.


src/amoc_config.erl
```diff
@@ -4,12 +4,13 @@
 %%==============================================================================
 -module(amoc_config).
 
--export([get/1,
-         get/2]).
-
--export([parse_scenario_settings/1,
+-export([get/1,get/2,
```
michalwski: Missing space after `,`
```suggestion
-export([get/1, get/2,
```

doc/distributed.md
```diff
@@ -25,23 +25,25 @@ docker run --rm -d --name=graphite --network amoc-test-network \
...
+    --network amoc-test-network \
     -e AMOC_HOSTS="\"amoc-1\",\"amoc-2\"" \
     -e AMOC_GRAPHITE_HOST=graphite \
     -e AMOC_GRAPHITE_PORT=2003 \
     -e AMOC_PREFIX=amoc1 \
-    --health-cmd="${PATH_TO_AMOC} status" \
+    --health-cmd="/home/amoc/amoc/bin/amoc status" \
     -p 8081:4000 \
-    amoc-reworked:latest
+    amoc_image:tag
```
michalwski: I think it'd be better to put a real amoc image tag like `mongooseim/amoc:latest`. People probably will want to copy and paste this command to start amoc node.

src/amoc_metrics.erl
```diff
@@ -75,16 +72,53 @@ metric_report_datapoints(gauge)     -> [value];
...
+
+maybe_add_reporter() ->
+    Reporter = get_reporter(),
+    case lists:keyfind(Reporter, 1, exometer_report:list_reporters()) of
+        {Reporter, _} -> ok;
+        _->
+            case amoc_config:get(graphite_host) of
+                undefined -> ok;
+                Host ->
+                    Prefix =amoc_config:get(graphite_prefix,net_adm:localhost()),
```
michalwski: Missing spaces after `=`
```suggestion
                    Prefix = amoc_config:get(graphite_prefix, net_adm:localhost()),
```

src/amoc_metrics.erl
```diff
@@ -75,16 +72,53 @@ metric_report_datapoints(gauge)     -> [value];
...
+                    Port = amoc_config:get(graphite_port,2003),
+                    Options = [{module, exometer_report_graphite},
+                               {prefix, Prefix},
+                               {host, Host},
+                               {port, Port},
+                               {api_key, ""}],
+                    exometer_report:add_reporter(Reporter,Options)
```
michalwski: Missing space
```suggestion
                    exometer_report:add_reporter(Reporter, Options)
```

src/amoc_metrics.erl
```diff
@@ -75,16 +72,53 @@ metric_report_datapoints(gauge)     -> [value];
...
+maybe_add_reporter() ->
+    Reporter = get_reporter(),
+    case lists:keyfind(Reporter, 1, exometer_report:list_reporters()) of
+        {Reporter, _} -> ok;
+        _->
+            case amoc_config:get(graphite_host) of
+                undefined -> ok;
+                Host ->
+                    Prefix =amoc_config:get(graphite_prefix,net_adm:localhost()),
+                    Port = amoc_config:get(graphite_port,2003),
```
michalwski: Missing space
```suggestion
                    Port = amoc_config:get(graphite_port, 2003),
```

src/amoc_metrics.erl
michalwski: Missing space
```suggestion
                    exometer_report:add_reporter(Reporter, Options)
```

src/amoc_metrics.erl
```diff
@@ -75,16 +72,53 @@ metric_report_datapoints(gauge)     -> [value];
...
+subsribe_default_metrics() ->
+    maybe_subscribe([amoc, users], [size]),
+    maybe_subscribe([erlang, system_info], [port_count, process_count]),
+    maybe_subscribe([erlang, memory], [total, processes, processes_used, system, binary, ets]),
+    maybe_subscribe([amoc, times, connection], metric_report_datapoints(histogram)),
+    maybe_subscribe([amoc, counters, connections], metric_report_datapoints(spiral)),
+    maybe_subscribe([amoc, counters, connection_failures], metric_report_datapoints(spiral)).
+
+get_reporter() ->
+    amoc_config:get(metrics_reporter, ?AMOC_DEFAULT_METRICS_REPORTER).
```
michalwski: Thanks for making the metrics reporter configurable. What will be returned here if there is `AMOC_metrics_reporter` env var set to, say, `my_own_reporter`. Will it be atom (as we expect it to be) or a string value?

doc/distributed.md
DenysGonchar: actually, `tag` is a valid tag. you can copy-past that. i did it for testing :)

src/amoc_metrics.erl
DenysGonchar: now it can be either `AMOC_metrics_reporter` or `AMOC_METRICS_REPORTER` os env. (however for the `2.0` i would prefer to restrict it to the "all capitals" option) or even `metrics_reporter` app env. configured in app.config

it must be atom, but the value of env variable is evaluated. so it should look like this `export AMOC_METRICS_REPORTER="some_atom"`

documentation for amoc_config is really good idea. it can contain API description, AMOC default options description and some reference how to pass scenario configurations properly.

doc/distributed.md
michalwski: How about `amoc_image:tag` this only works if you have it locally. If you don't then docker will attempt do download it from dockerhub and there, there is no `amoc_image:tag`.

michalwski: Ah, now I see. The image is build in a step at the very beginning of the doc page. No need to change that.

DenysGonchar: and the first instruction in the file :
```
docker build -f Dockerfile -t amoc_image:tag .
```
this PR is not about documentation, i just made instructions in the file compatible and copy-pastable.

src/amoc_metrics.erl
michalwski: I think we need a test for that case :)

DenysGonchar: in my example that is atom, but if you would have something like:
`export AMOC_SOME_VAR="[\"some_atom\"]"`
than it will be a list with one string element. the value of the os. env. is evaluated

PR #108, opened 2019-10-31
Use logger for logging from amoc

telezynski:
In this PR logger is used for logging from Amoc while dependencies still use lager.

### DIFF
--- a/priv/app.config
+++ b/priv/app.config
@@ -1,1 +1,17 @@
-[{ssl, [{session_cb, amoc_always_null_ssl_session_cache}]}].
+%% -*-erlang-*-
+
+[
+ {kernel, [
+           {logger_level, warning},
+           {logger,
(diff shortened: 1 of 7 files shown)


priv/app.config
```diff
@@ -1 +1,17 @@
-[{ssl, [{session_cb, amoc_always_null_ssl_session_cache}]}].
+%% -*-erlang-*-
+
+[
+ {kernel, [
+           {logger_level, warning},
```
michalwski: As far as I understand the default level is `notice`. I think we can assume that `notice` is also default for Amoc. This filters out all the progress reports form SASL wile allow us to print important information from scenarios on less significant and more suitable level than `warning` without changing the log level in runtime.

telezynski: I think it was a leftover from lager integration.

I removed it now, and checked the logs still do not print progress reports.

PR #109, opened 2019-10-31
[WIP] Amoc 2.0.0 cleanup

DenysGonchar:
Amoc 2.0 changes.

ready for review.


.dockerignore
```diff
@@ -1,20 +0,0 @@
-*.css
```
michalwski: We are still building docker images from Amoc code, and we'll continue to do so. Why exactly did you remove the file?

Dockerfile
```diff
@@ -29,14 +30,13 @@ MAINTAINER Erlang Solutions <mongoose-im@erlang-solutions.com>
 
 RUN useradd -ms /bin/bash amoc
 
-COPY --from=0 /amoc_build/_build/default/rel/amoc/ /home/amoc/amoc/
+COPY --from=amoc-build /amoc_build/_build/default/rel/amoc/ /home/amoc/amoc/
```
michalwski: That's clever!

src/amoc_always_null_ssl_session_cache.erl
```diff
@@ -1,42 +0,0 @@
--module(amoc_always_null_ssl_session_cache).
```
michalwski: Why do we remove this file? This was a workaround for high memory consumption in scenarios doing lots of HTTPs requests. Was this improved in recent Erlang/OTP versions?

src/amoc.app.src
```diff
@@ -29,10 +28,7 @@
                 []},
                {[amoc, users],
                 {function, ets, info, [amoc_users], proplist, [size]},
-                []},
-               {[amoc, times, connection], histogram, []},
-               {[amoc, counters, connections], spiral, []},
-               {[amoc, counters, connection_failures], spiral, []}
```
michalwski: These metrics are used by existing scenarios (in amoc-arsenal-xmpp). Are you going to move creation and subscription of these metrics to `amoc-arsenal-xmpp` in a linked PR? If no I'd rather leave them here.

src/amoc_networking.erl
```diff
@@ -1,37 +0,0 @@
--module(amoc_networking).
```
michalwski: I'm strongly against removing the module. It was (and may still be) used in case we have many network interfaces and want to use them all to open connections to the target node. I'd rather document this helper module with examples how to use it, instead of removing it.

.dockerignore
DenysGonchar: now it's `git clean -ffxd` executed before the build in the Dockerfile

src/amoc_always_null_ssl_session_cache.erl
DenysGonchar: this one is not used in docker container anyway. also it's better to move it into the `amoc-arsenal-xmpp`

src/amoc.app.src
DenysGonchar: yeap,  this one must be moved to amoc-arsenal-xmpp

src/amoc_networking.erl
DenysGonchar: the idea was to move it to `amoc-arsenal-xmpp`

michalwski:
The corresponding amoc-arsenal PR is in esl/amoc-arsenal-xmpp#11

PR #111, opened 2019-11-08
Distribute scenario to all amoc nodes

telezynski:


src/amoc_api_scenarios_handler.erl
```diff
@@ -128,21 +130,11 @@ to_json(Req0, State) ->
...
-                       ModuleName,
-                       ScenarioPath),
-            ResultBody = case Result of
-                             ok -> Result;
-                             {error, Errors, _Warnings} ->
-                                 R = io_lib:format("~p", [Errors]),
-                                 erlang:list_to_bitstring(lists:flatten(R))
-                         end,
+            %%Nodes = [amoc_slave:node_name(Node) || Node <- amoc_config:get(hosts, [])],
+            Nodes = [node()],
```
michalwski: The list of nodes consists only of local nodes. Do we want to extend it by other nodes amoc nodes? In the `amoc_dist` there is not exproted function `amoc_nodes`. Maybe we can export it and use it here?

src/amoc_api_scenarios_handler.erl
```diff
@@ -128,21 +130,11 @@ to_json(Req0, State) ->
...
+            io:format("**** Nodes: ~p~n", [Nodes]),
```
michalwski: Do we want to keep this debug information, if so it'd be better to change it to ?LOG_NOTICE or ?LOG_INFO

src/amoc_api_scenarios_handler.erl
telezynski: I was hesitating what to do.
I think the best would be to export `amoc_nodes` and have one place where we will control how we want to calculate it.

src/amoc_api_scenarios_handler.erl
telezynski: To be deleted

telezynski:
Added integration_test directory with scripts to build and run docker containers,
and also to run a test of installation of a scenario.
Test uses curl via HTTP API

PR #121, opened 2020-03-19
Update path to binary in local-run documentation

joesanford:
As the Makefile shows `make rel` resulting in a call to `rebar as prod tar`, the path of the resulting binary would be nested under the `prod` directory for the release.

### DIFF
--- a/doc/local-run.md
+++ b/doc/local-run.md
@@ -2,7 +2,7 @@
 
 Everything you need to do is to create the release. To achieve that run:
 `make rel`. Now you are ready to test our scenario locally with one Amoc node;
-to start the node run `_build/default/rel/amoc/bin/amoc console`.
+to start the node run `_build/default/prod/rel/amoc/bin/amoc console`.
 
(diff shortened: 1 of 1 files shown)


doc/local-run.md
```diff
@@ -2,7 +2,7 @@
 
 Everything you need to do is to create the release. To achieve that run:
 `make rel`. Now you are ready to test our scenario locally with one Amoc node;
-to start the node run `_build/default/rel/amoc/bin/amoc console`.
+to start the node run `_build/default/prod/rel/amoc/bin/amoc console`.
```
michalwski: The path is almost correct. I've just made a clean build (without anything in `_build` directory) and the correct path to amoc binary is:
```suggestion
to start the node run `_build/prod/rel/amoc/bin/amoc console`.
```

joesanford:
Looks like I had some mixups between latest and the older version we are on. Updated!

michalwski:
Thanks!

PR #123, opened 2020-05-15
Scenario upload rework

DenysGonchar:
This PR reworks scenario (module) uploading to amoc cluster. Now the PUT HTTP method is used instead of POST, so JSON encoding of erlang code is not required any more. 

Please note that there is no support for file uploading using PUT method in the current swagger 2.0 user interface.

REST API documentation will be updated in a subsequent PR.


michalwski:
Oh, and I just noticed, that dialyzer returns some errors after these changes.

DenysGonchar:
> Why do we need to change the method from POST to PUT? PUT seems better as file upload is idempotent and it's the default method in `curl -T`. Was this the reason?

yes, PUT is more appropriate for these purposes.

chrzaszcz:
I was still waiting for the answer for this question:

> I see that there is a new path /upload needed to handle the new content type, right? Why do we need application/json for /scenarios though as there is only GET there now?

DenysGonchar:
Because that is declaration of response type for this GET request.

chrzaszcz:
> Because that is declaration of response type for this GET request.

I still don't think I got the answer. Let me rephrase the question then:

Why use a separate path `/uploads` instead of `/scenarios` as before? It is not wrong, I just would like to know your exact motivation.

DenysGonchar:
I prefer to have different paths for different functionality. it's easier to keep codebase clean, and easier to explain in docs.

michalwski:
It just came to my mind that, maybe path like `/scenarios/upload` would more self-explanatory here.

PR #124, opened 2020-05-27
Rest api rework

DenysGonchar:
This PR replaces an old cowboy-swagger based REST API with the new openapi-generator implementation.

The generated server code is in the separate repo https://github.com/esl/amoc_rest

The whole migration process is described [here](https://github.com/esl/amoc/blob/7f73d73/open_api/README.md) in details

NB: it is important to merge this PR using "create a merge commit" option!!!


open_api/amoc_rest/rebar.config
```diff
@@ -0,0 +1,6 @@
+{deps, [
+    {cowboy, {git, "https://github.com/ninenines/cowboy.git", {tag, "2.5.0"}}},
```
michalwski, line 2: That's quite an old version of cowboy, cat it work with `2.7.0`?

DenysGonchar:
>     1. I can see that we have multiple static files describing the API
>        
>        * `open_api/amoc-openapi.yml`
>        * `open_api/amoc-swagger.json`
>        * `open_api/amoc_rest/priv/openapi.json`
>        
>        Can we have only one?
 - yes the whole open_api directory will be removed in the last commit, it is there just to keep the history of migration from amoc-swagger to openapi-generator in git. The only file we need is `amoc-openapi.yml`. there is also `priv/openapi.json` file - the generated json version of the yaml file, we should not edit this json file manually.

>     2. As far as I understand the approach taken in this PR generates the erlang code from the API spec. I find the other way around (the spec generated from the code) more convenient.
 - the main problem with that spec generation approach - it's almost impossible to keep everything up to date. generation of the spec is not ideal, it's hard to support JSON schemas and validation of input and output must be done manually. it's based on cowboy-trails project which is not well maintained, as well as cowboy-swagger. also both projects are purely documented. 

>     3. The auto-generated code doesn't fit our (any?) formatting rules
 - ideally, this generated code should be used as is, w/o any changes. It also can be moved out to the separate rebar3 project and used in amoc as a dependency, but from my perspective this is not necessary.

open_api/amoc_rest/rebar.config
DenysGonchar: i think we can safely upgrade it, will check that.

src/rest_api/amoc_api_logic_handler.erl
```diff
@@ -0,0 +1,84 @@
...
+                                           id := Resource }) ->
+    Scenario = binary_to_atom(Resource, latin1),
+    case amoc_api_scenario_status:test_status(Scenario) of
+        doesnt_exist ->
+            {404, #{}, #{}};
+        loaded ->
+            Users = maps:get(<<"users">>, Body),
+            SettingsMap = maps:get(<<"settings">>, Body, #{}),
+            Settings = [begin
+                            Key = binary_to_atom(K, latin1),
```
michalwski, line 58: This is probably a copy-paste. I think it'd be good to change `latin1` to `utf8` as we have in other projects (MongooseIM for instance). It'd be even better to use `binart_to_existing_atom`. The perfect solution would be to parse the binary and return atom only when the binary is acceptable.

src/rest_api/amoc_api_logic_handler.erl
```diff
@@ -0,0 +1,84 @@
...
+    FailedToConnect = maps:get(failed_to_connect, Status, []),
+    ConnectionLost = maps:get(connection_lost, Status, []),
+    Up = [{Node, <<"up">>} || Node <- [node() | Connected]],
+    DownNodes = lists:usort(FailedToConnect ++ ConnectionLost),
+    Down = [{Node, <<"down">>} || Node <- DownNodes],
+    ResponseList = Up ++ Down,
+    {200, #{}, [{<<"nodes">>, ResponseList}]};
+handle_request('ScenariosGet', _Req, _Context) ->
+    Scenarios = amoc_scenario:list_scenario_modules(),
+    BinaryScenarios = [atom_to_binary(S, latin1) || S <- Scenarios],
```
michalwski, line 30: Is there a reason behind the `latin1` encoding in `amoc`? Maybe `utf8` could be used here?

src/rest_api/amoc_api_logic_handler.erl
```diff
@@ -0,0 +1,84 @@
...
+    {200, #{}, [{<<"scenarios">>, BinaryScenarios}]};
+handle_request('StatusGet', _Req, _Context) ->
+    Apps = application:which_applications(),
+    Status = case lists:keyfind(amoc, 1, Apps) of
+                 {amoc, _Desc, _Vsn} -> <<"up">>;
+                 false -> <<"down">>
+             end,
+    {200, #{}, [{<<"node_status">>, Status}]};
+handle_request('ScenariosIdGet', _Req, #{id := Resource}) ->
+    Scenario = binary_to_atom(Resource, latin1),
```
michalwski, line 40: Denger! Please try to avoid `binary_to_atom`. Also, I think we can use the `utf8` encoding.

src/rest_api/amoc_api_logic_handler.erl
```diff
@@ -0,0 +1,84 @@
...
+    case amoc_api_scenario_status:test_status(Scenario) of
+        doesnt_exist ->
+            {404, #{}, #{}};
+        Status ->
+            BinStatus = atom_to_binary(Status, latin1),
+            {200, #{}, [{<<"scenario_status">>, BinStatus}]}
+    end;
+handle_request('ScenariosIdPatch', _Req, #{'ScenarioExecution' := Body,
+                                           id := Resource }) ->
+    Scenario = binary_to_atom(Resource, latin1),
```
michalwski, line 50: Denger! Please try to avoid `binary_to_atom`. Also, I think we can use the `utf8` encoding.

src/rest_api/amoc_api_logic_handler.erl
```diff
@@ -0,0 +1,84 @@
...
+    case amoc_api_upload_scenario:upload(ModuleSource) of
+        ok ->
+            {200, #{}, [{<<"compile">>, <<"ok">>}]};
+        {error, invalid_module} ->
+            {400, #{}, [{<<"error">>, <<"invalid module">>}]};
+        {error, Error} ->
+            {200, #{}, [{<<"compile">>, Error}]}
+    end;
+handle_request(OperationID, Req, Context) ->
+    error_logger:error_msg(
```
michalwski, line 80: In amoc we are using the built-in Erlang logger. I think it'd be better to be consistent with that and use it here as well.

src/rest_api/amoc_api_scenario_status.erl
```diff
@@ -0,0 +1,41 @@
+%%==============================================================================
+%% Copyright 2015 Erlang Solutions Ltd.
```
michalwski, line 2:
```suggestion
%% Copyright 2020 Erlang Solutions Ltd.
```

src/rest_api/amoc_api_upload_scenario.erl
```diff
@@ -0,0 +1,70 @@
+%%==============================================================================
+%% Copyright 2015 Erlang Solutions Ltd.
```
michalwski, line 2:
```suggestion
%% Copyright 2020 Erlang Solutions Ltd.
```

test/amoc_api_scenario_handler_SUITE.erl
```diff
@@ -59,17 +57,6 @@ end_per_testcase(
 end_per_testcase(_, _Config) ->
     destroy_env().
 
-get_scenario_status_returns_200_when_scenario_exists(_Config) ->
```
michalwski, line 62: Why this test case was removed?

test/amoc_api_status_handler_SUITE.erl
```diff
@@ -43,15 +31,6 @@ returns_up_when_amoc_up_online(_Config) ->
     ?assertEqual(200, CodeHttp),
     ?assertMatch({[{<<"node_status">>, <<"up">>}]}, Body).
 
-
-returns_down_when_api_up_and_amoc_down_offline(_Config) ->
```
michalwski, line 47: Why this test case was removed?

src/rest_api/amoc_api_logic_handler.erl
DenysGonchar: latin1 should be good enough for the module name, but I can change it to utf8 if you want.

src/rest_api/amoc_api_logic_handler.erl
DenysGonchar: yes, I tried to introduce changes step by step. so the whole implementation of the rest api handlers is extracted from the previous cowboy-swagger based implementation. the risk is quite low, taking into account that this API should be called just once during the whole amoc app lifecycle. And the goal of this PR is to have equivalent API after migration, while binart_to_existing_atom usage requires introduction of some new error code. Also, `amoc_config_env:parse_value(V)` can still introduce new atoms, but we definitely don't want to change that functionality.

src/rest_api/amoc_api_logic_handler.erl
DenysGonchar: sure, that is a copy-past from the default handler. thanks for catching that.

test/amoc_api_scenario_handler_SUITE.erl
DenysGonchar: it's a very old one. `given_test_status_mocked(true)` - this is completely invalid, the mocked function should never return `true` value (even before the migration), `loaded` is used instead.

test/amoc_api_status_handler_SUITE.erl
DenysGonchar: it doesn't introduce any value. it tests absolutely the same thing as`returns_down_when_api_up_and_amoc_down_online`, but restricts internal implementation.

src/rest_api/amoc_api_logic_handler.erl
DenysGonchar: it's extracted from the previous implementation. this api should be called just once during the whole lifecycle, so the risk is quite low. also `latin1` should be enough for the module name. but in this particular case we can replace it with `binart_to_existing_atom` in a relatively simple manner. will do it.

src/rest_api/amoc_api_logic_handler.erl
DenysGonchar: it's extracted from the previous implementation. this api should be called just once during the whole lifecycle, so the risk is quite low. also `latin1` should be enough for the module name. but in this particular case we can replace it with `binart_to_existing_atom` in a relatively simple manner. will do it.

PR #126, opened 2020-07-03
Elvis integration

DenysGonchar:
integration of CI elvis verification.
no functional changes, mostly whitespaces.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,11 @@
 sudo: false
 
 language: erlang
+
 otp_release:
     - 22.3
     - 21.3
+
 env:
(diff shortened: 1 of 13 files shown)


DenysGonchar:
> I've heard that elvis on CI was quite annoying and last thing I remember is that @michalwski was against it.

AFAIK The only thing he asked for is to do it in a separate PR

> Also, the rules regarding Elvis here are quite lightweight and conservative, they seem totally unproblematic to me, but I'm thinking, they're so so soft, what's the rationale behind not adding other rules?

these are only the changes to the default rules :)

PR #130, opened 2020-08-03
Scenario descr rest api

DenysGonchar:
Implementation of `/scenarios/{id}/info` and update of  `/scenarios/{id}` GET methods.
`/scenarios/{id}` is tested manually. I haven't extended tests for it since we are going to rework it in the next iteration.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,12 @@
-FROM phusion/baseimage as amoc-build
+FROM phusion/baseimage:18.04-1.0.0 as amoc-build
 
 ARG otp_vsn=22.3.4-1
 
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
-        git \
-        make \
-        wget && \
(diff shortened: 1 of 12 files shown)


Dockerfile
```diff
@@ -1,14 +1,12 @@
-FROM phusion/baseimage as amoc-build
+FROM phusion/baseimage:18.04-1.0.0 as amoc-build
```
NelsonVides, line 1: As in https://github.com/esl/amoc-arsenal-xmpp/pull/20, I'd actually change it to `bionic-1.0.0`. Checking it on github, they both actually point to the same commit hash, so they're equivalent, but the `bionic-` version is the one they have released as an "official" tag.

DenysGonchar: I would rather leave the YY.MM version, I usually do not remember which "animal" correspond to which ubuntu release. but now it's pretty clear how old the version is.

PR #131, opened 2020-08-04
REST API for scenario execution

chrzaszcz:
Follows the API definition: https://github.com/esl/amoc_rest/blob/1d615924/openapi.yaml

- Added and tested the new REST API endpoints
- Added support for `update_settings` in `amoc_dist`
- Fixed some minor issues with type specs

There is an increasing need to add tests for `amoc_dist` - this module is untested. This should be done in a separate issue.

See the commit messages below for more details.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -6,7 +6,7 @@
     {exometer_core, {git, "https://github.com/esl/exometer_core.git", {branch, "master"}}},
     {exometer_report_graphite, {git, "https://github.com/esl/exometer_report_graphite.git", {branch, "master"}}},
     %% when updating amoc_rest version, don't forget to update it at ./doc/http-api.md as well.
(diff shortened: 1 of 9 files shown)


rebar.config
```diff
@@ -6,7 +6,7 @@
     {exometer_core, {git, "https://github.com/esl/exometer_core.git", {branch, "master"}}},
     {exometer_report_graphite, {git, "https://github.com/esl/exometer_report_graphite.git", {branch, "master"}}},
     %% when updating amoc_rest version, don't forget to update it at ./doc/http-api.md as well.
-    {amoc_rest, {git, "https://github.com/esl/amoc_rest.git", {ref, "7a46d9f"}}},
+    {amoc_rest, {git, "https://github.com/esl/amoc_rest.git", {ref, "1d615924"}}},
```
DenysGonchar, line 9: please update the link in `./doc/http-api.md`

PR #132, opened 2020-08-10
Amoc config env cleanup

DenysGonchar:
removing possibility to provide configuration parameters values via erlang app env. variables.

this is a legacy functionality that rather increases the complexity than provides any benefits, it became confusing after splitting up amoc/scenarios/helpers into several erlang applications.

### DIFF
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -1,6 +1,6 @@
 ## Configuration
 
-Amoc is configured through the environment variables (uppercase with prefix ``AMOC_``).
+Amoc is configured through environment variables (uppercase with prefix ``AMOC_``).
 Note that the environment variables are evaluated as erlang code
 
 Amoc supports the following generic configuration parameters:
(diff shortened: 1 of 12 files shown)


doc/configuration.md
```diff
@@ -1,56 +1,48 @@
...
-    * app.config example:  `{api_port, 4000}`
+    * example: `AMOC_API_PORT="4000"`
                                       
 * ``interarrival`` - a delay (in ms, for each node in the cluster independently) between creating the processes
   for two consecutive users:
     * default value - 50 ms.
-    * os env example: `AMOC_INTERARRIVAL="50"`
-    * app.config example:  `{interarrival, 50}`
+    * example: `AMOC_INTERARRIVAL="50"`
+    * this parameter can be updated runtime (in the same way as scenario configuration).
```
chrzaszcz, line 20: at runtime

doc/configuration.md
```diff
@@ -1,56 +1,48 @@
 ## Configuration
 
-Amoc is configured through the OTP application environment variables that
-can be set in the `app.config` configuration file (see [rebar3 documentation](https://www.rebar3.org/docs/releases#section-overlays))
-or using the operating system environment variables (uppercase with prefix ``AMOC_``). Note that the operating system environment variables are evaluated as erlang code
+Amoc is configured through the environment variables (uppercase with prefix ``AMOC_``).
```
chrzaszcz, line 3: for me 'the' is not needed here (we do not know yet which variables they are)

doc/configuration.md
```diff
@@ -65,6 +57,7 @@ parameters required for your scenario, however every scenario must declare (usin
 `-required_variable(...)` attributes) all the required parameters in advance. For more
 information, see the example [scenario module](../integration_test/dummy_scenario.erl)
 
+Scenario configuration also can be set/updated runtime using REST API.
```
chrzaszcz, line 60: at runtime

doc/configuration.md
DenysGonchar: i think "runtime" can be used as a flat adverb, but I don't mind changing it.

NelsonVides: I'm afraid runtime is a noun and it needs the particle

PR #135, opened 2020-11-24
switching from TravisCI to GH Actions

DenysGonchar:
this PR fixes the following issues with GH Actions environment:
* missing TRAVIS_OTP_RELEASE env variable
* by default checkout action fetches just one exact commit, since amoc uses 'git' version (see rebar.config) rebar3 searches for the latest git tag and so requires the whole repo
* 8084 port is used by GH actions environment, so for integration testing amoc REST API ports are mapped at 888x range
* GH actions environment has no tty device


integration_test/helper.sh
```diff
@@ -111,6 +113,12 @@ function wait_for_cmd() {
     return 1
 }
 
+function get_amoc_logs() {
```
chrzaszcz, line 116: I don't see this function used anywhere, am I missing something?

.github/workflows/ci.yml
```diff
@@ -12,14 +12,18 @@ jobs:
     strategy:
       matrix:
         otp: ['22.3', '21.3']
```
chrzaszcz, line 14: Maybe we could use OTP 23? Not required now of course.

chrzaszcz (commented):
Great to see the GH actions in action :-)
I have a few comments, could you answer?

integration_test/helper.sh
DenysGonchar: I used it for troubleshooting issue at CI environment and I think it's good to have in helpers.sh just in case.

.github/workflows/ci.yml
DenysGonchar: right now we cannot switch to OTP. [pg2](http://erlang.org/doc/man/pg2.html) is deprecated in OTP23 and xref check is failing, the rest checks are passing well w/o any issues. 

So we need to first switch from [pg2](http://erlang.org/doc/man/pg2.html) to something else, but if we switch to [pg](http://erlang.org/doc/man/pg.html) as OTP documentation suggests we limit ourselves to OTP23 and higher version since it's introduced only in OTP23.

chrzaszcz (approved):
Good!

PR #136, opened 2021-05-12
Otp24

NelsonVides:
~Waiting for https://github.com/esl/amoc_rest/pull/8/~


gustawlippa:
The tests are run for OTP 21.3 but `pg` was introduced in OTP 23.0.

NelsonVides:
> The tests are run for OTP 21.3 but `pg` was introduced in OTP 23.0. This is a welcome change but it should somehow work with the CI.

And that explains the failure, but that makes it all a bit more complicated, because pg2 was completely removed with otp24, which means we could only support the last two major releases, or not support the last. Or do some hacks to compile different things with different versions 😕 

@DenysGonchar

DenysGonchar:
yeah, that is known issues, and that's exactly the reason why we didn't have CI testing against OTP23. pg2 became deprecated since that release, but we didn't want to relax dialyzer/xref verification rules.

DenysGonchar:
I guess  now, having 2 major OTP release with pg module, we can replace pg2 with pg. 
Of course, we would have to make a new release with the comment that only OTP23 and OTP24 are supported and those who need to use pre OTP23 version can still use the previous Amoc release.

test/amoc_api_scenarios_handler_SUITE.erl
```diff
@@ -71,6 +71,19 @@ put_scenarios_returns_400_and_error_when_scenario_is_not_valid(_Config) ->
     ?assertEqual(400, CodeHttp),
     ?assertEqual(#{<<"error">> => <<"invalid module">>}, Body).
 
+-ifdef(OTP_RELEASE).
```
DenysGonchar, line 74: Please change to the following sequence:
```
-define(PARSE_ERROR,...).
-ifdef(OTP_RELEASE). %% being honest i don't think that we need this one
-if(?OTP_RELEASE >= 24).
-undefine(PARSE_ERROR).
-define(PARSE_ERROR,...).
-endif.
-endif.
```

src/amoc_throttle/amoc_throttle_controller.erl
```diff
@@ -181,7 +181,6 @@ handle_call({change_rate_gradually, Name, LowRate, HighRate,
 handle_call({stop, Name}, _From, State) ->
     case all_processes(Name, stop) of
         ok ->
-            pg2:delete(Name),
```
DenysGonchar, line 184: is group removed in this case? can we recreate the group with the same name after this?

NelsonVides: https://github.com/erlang/otp/blob/4d2a17302a79a86b6c74c769b9fafd9fbd9be516/lib/kernel/src/pg.erl#L21-L53 Here's an explanation, non-existent and empty groups are treated the same in this new implementation, the reason being agreement in the case of coming back from a network split.

NelsonVides:
Github actions have some issue with nifs for a wxWidget dep not being loaded, no idea why, but that doesn't happen locally on my system. Needs investigation.

test/amoc_api_scenarios_handler_SUITE.erl
```diff
@@ -71,6 +71,14 @@ put_scenarios_returns_400_and_error_when_scenario_is_not_valid(_Config) ->
...
     ?assertEqual(#{<<"error">> => <<"invalid module">>}, Body).
 
+-define(PARSE_ERROR,
+        <<"\n                      [{2,erl_parse,[\"syntax error before: \",[]]}]}]\n">>).
+-if(?OTP_RELEASE >= 24).
+-undef(PARSE_ERROR).
+-define(PARSE_ERROR,
+        <<"\n                      [{{2,1},erl_parse,[\"syntax error before: \",[]]}]}]\n">>).
+-endif.
+
```
chrzaszcz, line 81:
```suggestion
-if(?OTP_RELEASE >= 24).
-define(PARSE_ERROR,
        <<"\n                      [{{2,1},erl_parse,[\"syntax error before: \",[]]}]}]\n">>).
-else.
-define(PARSE_ERROR,
        <<"\n                      [{2,erl_parse,[\"syntax error before: \",[]]}]}]\n">>).
-endif.
```

chrzaszcz (approved):
Looks good! There is one minor remark from me.

test/amoc_api_scenarios_handler_SUITE.erl
NelsonVides: I had something like that but Denys asked for it to be as it is now: https://github.com/esl/amoc/pull/136#discussion_r642712062

chrzaszcz: Any motivation @DenysGonchar? It's less readable.

DenysGonchar (approved):
It was surprisingly hard, but I really like the changes.
MB we should make a follow-up ticket to add a note regarding `{debug_info, keep}` relx flag in amoc documentation. I think it should be added [here](https://esl.github.io/amoc/doc/configuration.html), near to `-required_variable(...)` attribute explanation.

test/amoc_api_scenarios_handler_SUITE.erl
DenysGonchar: Being honest, I don't remember, and now it's hard to tell what triggered me w/o original commit. Your version is fine and indeed it's more readable.

PR #138, opened 2021-07-17
Fixing docsh dependency

DenysGonchar:
minor fixes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,19 @@
-FROM phusion/baseimage:focal-1.0.0 as amoc-build
+FROM phusion/baseimage:focal-1.0.0 as base-image
+FROM base-image as amoc-build
 
-RUN apt-get update && \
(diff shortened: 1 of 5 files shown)


vkatsuba (approved):

NelsonVides:
> minor fixes

Why? 🤔

DenysGonchar:
> > minor fixes
> 
> 
> 
> Why? 🤔 

Because when we use amoc (or any other app) as a dependency, we expect it to have a proper list of its own dependencies. Previously we were only including docsh in amoc release, which is now can be used solely for demo purposes. But amoc-arsenal-release didn't have it, and attempts to extract edoc during scenario/info REST calls were crashing.

BTW, there are also some issues in parsing edoc for some uploaded scenarios, but I haven't investigated that yet.

NelsonVides (approved):
LGTM

PR #140, opened 2021-12-28
Minor updates

DenysGonchar:
This PR fixes the default make target for amoc, so it can be built as `erlang.mk` dependency.
Also integration tests reworked a bit, now the testing cluster is managed with docker-compose command.


arcusfelis (approved):
ok

chrzaszcz:
Looks good!

PR #142, opened 2023-05-08
amoc_scenarios ets table improvement

NelsonVides:
Use a proper ETS lookup instead of ranging over all the table entries.


DenysGonchar (approved):
👌

PR #143, opened 2023-05-15
Update OTP, rebar, and ubuntu versions

NelsonVides:


chrzaszcz (approved):
Looks good, thanks for updating the deps 🙂

PR #144, opened 2023-05-15
Let amoc user processes exit normally

NelsonVides:
As discussed in the internal communication channel, trying to report `normal_user_stop` still results in an abnormal exit, so this is buggy code, and altogether makes processes promoted to OTP `gen` processes to die with terrible crash reports. Logic is “call start, if it crashes, crash too”, so it makes sense to replace the `throw` in that `amoc_user:stop/1` function with `exit(shutdown)` and not have any `try ... catch`.


chrzaszcz (approved):
Looks good 👍

PR #145, opened 2023-05-26
Amoc core

NelsonVides:
This PR is a proposal for a big change in how amoc works. Here, we separate metrics and APIs from what is actually the load-testing tooling, so that amoc itself can be a bare-bones dependency to be added to any project that wants to do load-testing.
- Amoc remain simply instrumented with telemetry, so that users can choose their own favourite metrics and logging systems by simply subscribing to the exposed telemetry events, without imposing graphite or prometheus.
- Likewise the API is removed, so that the user can write his own, be it REST or GraphQL (which we can write on our own as well and add them all as dependencies as desired.

TODOs so that this proposal is comprehensive and makes sense:
- [x] Move all removed REST code into `amoc-arsenal-xmpp` and have it possible to install `amoc` and `amoc_rest` and have just exactly the same functionality.
- [x] Move all removed `exometer` code into `amoc-arsenal-xmpp` and have it possible to install `exometer` and `exometer_report_graphite` and have just exactly the same functionality.


chrzaszcz (commented):
Looks good in general, I added some minor comments.

chrzaszcz (approved):
:+1:

PR #147, opened 2023-07-18
Hex package preparations

NelsonVides:
Throttle and coordinator documentation was moved to their respective modules, which reduces the redundancy of the documentation as specs for functions that also have specs.

Because ex_doc will clear and output its produce to the `doc/` dir, files from such dir have been moved to `guides/`.

The most controversial change could be the last commit, see commit message for reasonale.

---
To build locally you can fetch the branch and run
```
rebar3 ex_doc
```
And see in your browser `amoc/doc/index.html`.

How it looks locally:
image


test/amoc_config_scenario_SUITE.erl
```diff
@@ -16,9 +16,9 @@
 -required_variable(#{name => var1, description => "var1"}).
 -required_variable(#{name => var2, description => "var2", default_value => def2}).
 -override_variable(#{name => var2, description => "var2", default_value => val2,
-                     verification => [def2, val2], update => fun ?MODULE:update_fn/2}).
+                     verification => [def2, val2], update => {?MODULE, update_fn, 2}}).
```
DenysGonchar, line 19: I think we have to test both cases here, MFA tuple and full function reference.

test/amoc_config_attributes_SUITE.erl
```diff
@@ -20,15 +20,15 @@
     #{name => var3, description => "var3", default_value => def3,
       verification => [def3, another_atom]},
     #{name => var4, description => "var4", default_value => def4,
-      verification => fun ?MODULE:verify_value/1}
+      verification => {?MODULE, verify_value, 1}}
```
DenysGonchar, line 23: I think we have to test both cases here, MFA tuple and full function reference.

src/amoc_throttle/amoc_throttle_process.erl
```diff
@@ -117,11 +116,13 @@ handle_continue(maybe_run_fn, State) ->
     NewState = maybe_run_fn(State),
     {noreply, NewState, timeout(NewState)}.
 
+-spec format_status(term(), term()) -> term().
```
DenysGonchar, line 119: Is it anyhow better than automatically generated spec?

integration_test/dummy_scenario.erl
```diff
@@ -9,14 +9,13 @@
...
-    %% verification function must be supplied as a function pointer in a
-    %% 'fun module:function/arity' format. Note that it must be an exported function.
-    %% usage of 'fun function/arity' format in the module attribute simply will
-    %% not pass compilation.
+    %% verification function must be supplied as an mfa of the format
+    %% '{module, function, arity}' format. Note that it must be an exported function.
+    %% otherwise it will not pass compilation.
     #{name => var2, description => "description2", default_value => def2,
-      verification => fun ?MODULE:test_verification_function/1},
+      verification => {?MODULE, test_verification_function, 1}},
```
DenysGonchar, line 16: I think we have to test both cases here, MFA tuple and full function reference.

integration_test/dummy_scenario.erl
```diff
@@ -9,14 +9,13 @@
 
 -required_variable([
     %% var2 must be positively verified by the test_verification_function/1 function.
-    %% verification function must be supplied as a function pointer in a
-    %% 'fun module:function/arity' format. Note that it must be an exported function.
-    %% usage of 'fun function/arity' format in the module attribute simply will
-    %% not pass compilation.
+    %% verification function must be supplied as an mfa of the format
```
DenysGonchar, line 12: This comment is incorrect, because there are 2 formats supported now.

src/amoc_coordinator/amoc_coordinator.erl
```diff
@@ -1,6 +1,143 @@
 %%==============================================================================
-%% Copyright 2023 Erlang Solutions Ltd.
-%% Licensed under the Apache License, Version 2.0 (see LICENSE file)
+%% @doc This module allows to synchronize the users and act on groups of them.
```
DenysGonchar, line 2: Being honest, I'm not a big fun of such big documentation blocks in the header. I would rather keep a separated markdown doc (just maybe move it into the `guides` directory, so all the documentation is at one place).

src/amoc_throttle/amoc_throttle.erl
```diff
@@ -1,6 +1,148 @@
 %%==============================================================================
-%% Copyright 2023 Erlang Solutions Ltd.
-%% Licensed under the Apache License, Version 2.0 (see LICENSE file)
+%% @doc This module allows to synchronize the users and act on groups of them.
```
DenysGonchar, line 2: Being honest, I'm not a big fun of such big documentation blocks in the header. I would rather keep a separated markdown doc (just maybe move it into the `guides` directory, so all the documentation is at one place).

src/amoc_coordinator/amoc_coordinator.erl
NelsonVides: I have an idea then.

src/amoc_throttle/amoc_throttle_process.erl
DenysGonchar: is it possible to exclude that function from ex_doc, and the whole file as well?  it is an internal module.

PR #153, opened 2023-10-23
Add optional process to monitor amoc users

aleklisi:
The amoc users monitor is a process to keep track of what's happening with amoc users. It addresses my issue of not being able to easily determine if load/stress test is successful or users are failing, without digging into logs.


NelsonVides:
Hey @aleklisi!

I'd say this is not the right way to do this, it introduces yet another layer of abstraction to a repo that is supposed to be nothing more than a process orchestrator and scenario administration. Also gives a third process that will be notified of a user termination — `amoc_controller` and `amoc_users_sup` already both get this and that is something I'm planning to remove, a ton of users dying at once would flood two process with down messages, I don't want a third here.

Note that when #162 is merged, you could instead subscribe a handler to `[amoc, scenario, user, stop | exception]` as `telemetry:span/3` raises. These run in the user process, so an exit kill would prevent them from running, but kills should only come from a system going down anyway.

aleklisi:
@NelsonVides, thanks for the reply.

This optional process is provided for extra convenience when tracking how simulated users are progressing or not and it is meant to be used only when developing new stress test scenarios or working on somewhat unstable stress tests with eg small number of users failing in some edge cases and not to monitor all users in a full-scale scenario.

PR #155, opened 2023-11-27
Telemetry curation

NelsonVides:
I'm going to want timestamps for all these telemetry events, so that metrics collectors can aggregate and draw plots based on time.

I also want an event for scenario start and stop which will mimic telemetry's `span` mechanism, but which here it is done in this way because the controller separately calls the beginning and the end of the scenario and `span` wouldn't fit into this mechanism.

I also don't want logs to be raised by `amoc`. Logging might be desired to work differently, so instead of logs I'll also raise here telemetry events for internals, and in the subscribers I'll decide if I want to log something or not and how.


NelsonVides:
PR split into #162 and #163

PR #158, opened 2023-12-01
Permit scenario start when the previous is finished too

NelsonVides:
This happens when I run a scenario, this one finishes, and the status of the controller is left as `finished`, and then I want to run a new scenario but the controller complains that the status is invalid, when really I could perfectly start a new one.

Note that the controller has no tests on its own and therefore none are added here, actual tests want to be added later in a special PR.


arcusfelis (approved):
ok

PR #162, opened 2023-12-02
Telemetry/timestamps and spans

NelsonVides:
I'm going to want timestamps for all these telemetry events, so that metrics collectors can aggregate and draw plots based on time.

I also want an event for scenario start and stop which will mimic telemetry's span mechanism, but which here it is done in this way because the controller separately calls the beginning and the end of the scenario and span wouldn't fit into this mechanism.

Split from #155

### DIFF
--- a/guides/telemetry.md
+++ b/guides/telemetry.md
@@ -6,15 +6,15 @@ A telemetry span of a full scenario execution globally (i.e. the exported `init/
 ```erlang
 event_name: [amoc, scenario, run, _]
 measurements: #{} %% As described in `telemetry:span/3`
-metadata: #{scenario => module()} %% Plus as described in `telemetry:span/3`
+metadata: #{scenario := module()} %% Plus as described in `telemetry:span/3`
 ```
 
 A telemetry span of a full scenario execution for a user (i.e. the exported `start/1,2` function):
 ```erlang
 event_name: [amoc, scenario, user, _]
 measurements: #{} %% As described in `telemetry:span/3`
-metadata: #{scenario => module(),
-            user_id => non_neg_integer()} %% Plus as described in `telemetry:span/3`
+metadata: #{scenario := module(),
+            user_id := non_neg_integer()} %% Plus as described in `telemetry:span/3`
 ```
 
 ## Controller
@@ -22,8 +22,8 @@ metadata: #{scenario => module(),
 Indicates the number of users manually added or removed
 ```erlang
 event_name: [amoc, controller, users]
-measurements: #{count => non_neg_integer()}
-metadata: #{monotonic_time => integer(), scenario => module(), type => add | remove}
+measurements: #{count := non_neg_integer()}
+metadata: #{monotonic_time := integer(), scenario := module(), type := add | remove}
 ```
 
 ## Throttle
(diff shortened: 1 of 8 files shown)


src/amoc_telemetry.erl
```diff
@@ -0,0 +1,15 @@
+%% @private
+%% @copyright 2023 Erlang Solutions Ltd.
+-module(amoc_telemetry).
+
+-export([execute/3]).
+
+-spec execute(EventName, Measurements, Metadata) -> ok when
+      EventName :: telemetry:event_name(),
+      Measurements :: telemetry:event_measurements() | telemetry:event_value(),
```
DenysGonchar, line 9: isn't `event_value()` deprecated?

NelsonVides: True, I was just copying specs from telemetry. Will remove.

guides/telemetry.md
```diff
@@ -2,63 +2,69 @@ Amoc also exposes the following telemetry events:
...
-metadata: #{name => atom()}
+measurements: #{rate := non_neg_integer()}
+metadata: #{monotonic_time := integer(), name := atom(), msg => binary()}
 ```
 
 ### Request
 
 Raised when a process client requests to be allowed pass through a throttled mechanism.
-
+This event is raised only on the master node.
```
DenysGonchar, line 53: this is incorrect, master reports overall requests/executions. worker nodes report requests/executions that they issue.

NelsonVides: I believed throttle processes are started only on the master node 🤔

DenysGonchar: yes they are. but requests are issues on the worker node, and execution is done on the worker node as well. so we report this metric twice, from the trottle process perspective (on the master node), and on request/execution (on the worker node). and also we ensure there's no duplication of reporting for a single node execution.

guides/telemetry.md
```diff
@@ -2,63 +2,67 @@ Amoc also exposes the following telemetry events:
 
 ## Scenario
 
-A telemetry span of a full scenario execution
+A telemetry span of a full scenario execution globally (i.e. the exported `init/0` function):
```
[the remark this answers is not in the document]
NelsonVides: Agree to everything except the idea of running `Scenario:start/` as a telemetry span in the `amoc_scenario` module, as this will run in the controller process and measure the time it takes to spawn an erlang process only, instead of measuring the lifespan of the entire amoc user. The current span must stay where it is already and it is the most useful one.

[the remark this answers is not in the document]
DenysGonchar: I'm not sure what you mean here, I propose to measure exactly scenario callback time:
https://github.com/esl/amoc/blob/master/src/amoc_scenario.erl#L64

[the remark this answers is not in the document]
DenysGonchar: well, all of the callbacks.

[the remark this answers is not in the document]
DenysGonchar: and amoc_scenario module works as a synchronous wrapper for the callbacks. it doesn't spawn anything.

src/amoc_scenario.erl
```diff
@@ -56,25 +61,39 @@ terminate(Scenario, State) ->
...
-          erlang:function_exported(Scenario, start, 1)} of
-        {true, _} ->
-            Scenario:start(Id, State);
-        {_, true} ->
-            Scenario:start(Id);
-        {false, false} ->
-            error("the scenario module must export either start/2 or start/1 function")
-    end.
+    Metadata = #{scenario => Scenario, state => State, user_id => Id},
+    Span = case {erlang:function_exported(Scenario, start, 2),
```
DenysGonchar, line 69: span must be a function

DenysGonchar (approved):
great PR, thanks for addressing all the comments.

PR #170, opened 2023-12-16
Controller/tests

NelsonVides:


test/controller_SUITE.erl
```diff
@@ -0,0 +1,181 @@
...
+
+remove_users_scenario_not_started_fails(_) ->
+    Ret = amoc_controller:add_users(1, 10),
+    ?assertMatch({error, {invalid_status, idle}}, Ret).
+
+check_status_with_running_users_is_correct(_) ->
+    do_start_scenario(testing_scenario, regular_vars()),
+    amoc_controller:add_users(1, 10),
+    Status = amoc_controller:get_status(),
+    ?assertMatch({running, testing_scenario, _, 10}, Status).
```
DenysGonchar, line 109: think we should check the number of users as well:
```
?assertEqual({running, testing_scenario, 10, 10}, Status)
```
 and it's better to use range starting from some other value than 1:
 ```
    amoc_controller:add_users(6, 10),
    Status = amoc_controller:get_status(),
    timer:sleep(50*6), % enough time to start 6 users with 50ms interarrival
    ?assertEqual({running, testing_scenario, 5, 10}, Status).
 ```

test/controller_SUITE.erl
```diff
@@ -0,0 +1,181 @@
...
+
+regular_vars_with_state() ->
+    [{interarrival, 1}, {testing_state_var1, def1}].
+
+other_vars_to_keep_quiet() ->
+    [{config_scenario_var1, unused_value}].
+
+wait_until_scenario_has_users(Scenario, Current, Total) ->
+    WaitUntilFun = fun amoc_controller:get_status/0,
+    WaitUntilValue = {running, Scenario, Current, Total},
```
DenysGonchar, line 180: that's not total, but the highest user ID used by the controller

test/controller_SUITE.erl
```diff
@@ -104,9 +104,10 @@ remove_users_scenario_not_started_fails(_) ->
 
 check_status_with_running_users_is_correct(_) ->
     do_start_scenario(testing_scenario, regular_vars()),
-    amoc_controller:add_users(1, 10),
-    Status = amoc_controller:get_status(),
-    ?assertMatch({running, testing_scenario, _, 10}, Status).
+    StartId = 6,
+    EndId = 10,
+    amoc_controller:add_users(6, 10),
```
DenysGonchar, line 109: why not `amoc_controller:add_users(StartId, EndId),`
anyway, that's not blocking.

DenysGonchar (approved):

PR #171, opened 2023-12-16
Controller/Return a more explicit running status from

NelsonVides:
This way the map has keys with proper names for the fields we're looking at.


jacekwegr (approved):
Looks good to me! 👍

PR #172, opened 2023-12-16
Controller/Deduplicate user supervision

NelsonVides:
In the original implementation, the process is monitored by two supervisors, the original amoc_users_sup, and the amoc_controller one. But this means that bursts of ups and downs overflow the mailbox of _two_ processes, one of them being actually the very critical controller.

With this reimplementation, only the amoc_users_sup tracks the processes, and all requests from the controller are asynchronous. This ensures the controller does not get blocked and can remain responsive to control requests.


NelsonVides:
Done in #175

PR #173, opened 2023-12-16
Controller/Support interarrival of zero

NelsonVides:
Useful if for example we want to spawn a big pool of users all at once. Optimised by not setting any timers, but simply looping new user requests.


NelsonVides:
Done in #175

PR #174, opened 2023-12-16
Accept ranges of users in `amoc_coordinator`

NelsonVides:
Sometimes we want to add some variance to the test and we want the coordinator to choose groups of roughly N users instead of uniformly choosing all groups to be of the exact same size. For this, we could give the coordinator a range of the type {Min, Max} were every group will be chosen to have N in the interval [Min, Max]. `rand:uniform/1` can be used to ensure that the distribution will be pretty uniform, and in a sufficiently large scale, converge to N.

As a good to have, the coordinator could keep statistics of the distribution of N in the interval.


src/coordinator/amoc_coordinator_worker.erl
```diff
@@ -125,6 +130,12 @@ safe_executions(Fun, Args) ->
         _:_ -> ok
     end.
 
+-spec calculate_n(amoc_coordinator:num_of_users()) -> all | pos_integer().
+calculate_n({Min, Diff}) ->
```
arcusfelis, line 134: maybe could keep `{Min, Max}` format everywhere, not two formats

arcusfelis (approved):
looks ok

PR #175, opened 2023-12-16
Users supervision tree

NelsonVides:
## Dedup controller and users sup #172 
In the original implementation, the process is monitored by two supervisors, the original amoc_users_sup, and the amoc_controller one. But this means that bursts of ups and downs overflow the mailbox of two processes, one of them being actually the very critical controller.

With this reimplementation, only the amoc_users_sup tracks the processes, and all requests from the controller are asynchronous. This ensures the controller does not get blocked and can remain responsive to control requests.

## Interarrival of zero #173 
Useful if for example we want to spawn a big pool of users all at once. Optimised by not setting any timers, but simply looping new user requests.

## Pool user supervision trees
If users go up and down too fast, amoc_users_sup becomes a bottleneck as it
can’t keep up with the requests. Improve the performance by creating a pool of
supervisors whose size is proportional to the number of schedulers, and ensure
distribution of users among them by for example hashing on the user_id.

We do this by creating a pool of `amoc_users_worker_sup` supervisors and having the whole API through `amoc_users_sup` choose supervisors according to requests.


src/users/amoc_users_worker_sup.erl
```diff
@@ -0,0 +1,138 @@
...
+%% @private
+-spec start_link(non_neg_integer()) -> {ok, pid()}.
+start_link(N) ->
+    gen_server:start_link(?MODULE, N, []).
+
+%% @private
+-spec init(non_neg_integer()) -> {ok, term()}.
+init(N) ->
+    process_flag(trap_exit, true),
+    Tid = ets:new(?MODULE,  [ordered_set, private]),
```
DenysGonchar, line 38: any particular reason to make it private? why not protected and not named?
these settings make troubleshooting much harder.

src/users/amoc_users_sup.erl
```diff
@@ -1,60 +1,167 @@
...
+        0 ->
+            amoc_controller:zero_users_running();
+        _ ->
+            ok
+    end.
+
+-spec start_child(amoc:scenario(), amoc_scenario:user_id(), any()) -> ok.
+start_child(Scenario, Id, ScenarioState) ->
+    Sup = get_sup_for_user_id(Id),
+    gen_server:cast(Sup, {start_child, Scenario, Id, ScenarioState}).
```
DenysGonchar, line 95: never expose details of internal `gen_server` protocol, create interface at `amoc_users_worker_sup` module and call it instead.

src/users/amoc_users_sup.erl
```diff
@@ -1,60 +1,167 @@
...
+%% group at once. This way we reduce the number of casts to each worker to always one, instead of
+%% depending on the number of users.
+-spec start_children(amoc:scenario(), [amoc_scenario:user_id()], any()) -> ok.
+start_children(Scenario, UserIds, ScenarioState) ->
+    KeyFun = fun(UserId) ->
+                     get_sup_for_user_id(UserId)
+             end,
+    Assignments = maps:groups_from_list(KeyFun, UserIds),
+    CastFun = fun (Sup, Users) ->
+                      gen_server:cast(Sup, {start_children, Scenario, Users, ScenarioState})
```
DenysGonchar, line 107: never expose details of internal `gen_server` protocol, create interface at `amoc_users_worker_sup` module and call it instead.

src/users/amoc_users_sup.erl
```diff
@@ -1,60 +1,167 @@
...
+              end,
+    maps:foreach(CastFun, Assignments).
+
+%% Assign a count of children each worker needs to remove
+%% in order to load-balance the request among all workers.
+-spec stop_children(non_neg_integer(), boolean()) -> non_neg_integer().
+stop_children(Count, Force) ->
+    {CountRemove, Assignments} = assign_counts(Count),
+    [ gen_server:cast(Sup, {stop_children, Int, Force}) || {Sup, Int} <- Assignments ],
```
DenysGonchar, line 116: never expose details of internal `gen_server` protocol, create interface at `amoc_users_worker_sup` module and call it instead.

src/users/amoc_users_sup.erl
```diff
@@ -1,60 +1,167 @@
...
+    CountRemove.
+
+-spec get_all_children() -> [{pid(), amoc_scenario:user_id()}].
+get_all_children() ->
+    #storage{sups = Sups} = persistent_term:get(?MODULE),
+    All = [ gen_server:call(Sup, get_all_children, infinity) || Sup <- tuple_to_list(Sups) ],
```
DenysGonchar, line 122: never expose details of internal `gen_server` protocol, create interface at `amoc_users_worker_sup` module and call it instead.

src/users/amoc_users_sup.erl
```diff
@@ -1,60 +1,167 @@
...
+    lists:flatten(All).
+
+-spec terminate_all_children() -> any().
+terminate_all_children() ->
+    #storage{sups = Sups} = persistent_term:get(?MODULE),
+    [ gen_server:cast(Sup, terminate_all_children) || Sup <- tuple_to_list(Sups) ].
```
DenysGonchar, line 128: never expose details of internal `gen_server` protocol, create interface at `amoc_users_worker_sup` module and call it instead.

src/users/amoc_users_worker_sup.erl
```diff
@@ -0,0 +1,138 @@
...
+%% @private
+-spec handle_call(any(), any(), state()) -> {reply, term(), state()}.
+handle_call(get_all_children, _From, #state{tid = Tid} = State) ->
+    Children = ets:tab2list(Tid),
+    {reply, Children, State}.
+
+%% @private
+-spec handle_cast(any(), state()) -> {noreply, state()}.
+handle_cast({start_child, Scenario, Id, ScenarioState}, #state{index = N, tid = Tid} = State) ->
+    case amoc_user:start_link(Scenario, Id, ScenarioState) of
```
DenysGonchar, line 50: please move this logic into a `start_child` function and reuse it for `start_children` processing

DenysGonchar (changes requested):
this new implementation sufferrs from the same issues as the existing one. removal operation is async and if we call `amoc_controller:remove_users/2` two times in a row, it may select the same users for removal. so it won't remove the expected amount of users (the number returned by `amoc_controller:remove_users/2`). I don't think that we need to worry about it, but we must to add a note in the documentation about this behaviour.

src/users/amoc_users_sup.erl
```diff
@@ -1,60 +1,167 @@
...
-    {ok, {SupFlags, [AChild]}}.
+               modules => [amoc_users_worker_sup]
+              }
+             || N <- lists:seq(1, erlang:system_info(schedulers_online)) ],
+    Strategy = #{strategy => one_for_one, intensity => 0},
+    {ok, {Strategy, Specs}}.
+
+%% API
+-spec init_storage() -> any().
+init_storage() ->
```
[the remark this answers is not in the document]
NelsonVides: The supervisor docs say that
> Notice that this function can also be called as a part of a code upgrade procedure. Therefore, the function is not to have any side effects. For more information about code upgrade of supervisors, see section [Changing a Supervisor](https://www.erlang.org/doc/design_principles/appup_cookbook#sup) in OTP Design Principles.

That's why I chose to put it aside. However we might actually never consider upgrades a thing for a load-test though 😄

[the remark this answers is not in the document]
NelsonVides: Oh, also, the other issue is that the supervisor hasn't started its children, so we don't know yet the pids to store in a pt.

src/amoc_controller.erl
```diff
@@ -279,33 +295,26 @@ handle_disable(#state{status = Status} = State) ->
...
-        _ ->
-            State
-    end.
+-spec handle_start_all_users(state()) -> state().
+handle_start_all_users(#state{create_users   = AllUsers,
+                              scenario       = Scenario,
+                              scenario_state = ScenarioState,
+                              tref = TRef} = State) ->
+    amoc_users_sup:start_children(Scenario, AllUsers, ScenarioState),
+    State#state{create_users = [], tref = maybe_stop_timer(TRef)}.
```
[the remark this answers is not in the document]
NelsonVides: Yeah, I'm already calling it precisely 😄

src/users/amoc_users_sup.erl
```diff
@@ -1,60 +1,167 @@
...
+-spec incr_no_of_users(non_neg_integer()) -> any().
+incr_no_of_users(SupNum) ->
+    #storage{user_count = Atomics} = persistent_term:get(?MODULE),
+    atomics:add(Atomics, 1, 1),
+    atomics:add(Atomics, SupNum + 1, 1).
+
+-spec decr_no_of_users(non_neg_integer()) -> ok.
+decr_no_of_users(SupNum) ->
+    #storage{user_count = Atomics} = persistent_term:get(?MODULE),
+    atomics:sub(Atomics, SupNum + 1, 1),
```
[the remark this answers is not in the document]
NelsonVides: Yeah I know, but I'm not using the result on this operation, this is just subtracting one on the index of the supervisor, it is the operation just below the one that checks if it is the total index the one being counted.

src/users/amoc_users_sup.erl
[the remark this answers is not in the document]
DenysGonchar: `start_link/0` calls `supervisor:start_link/3` and `supervisor:start_link/3` returns only after sup. tree initialization is done. so by the time of returning from that function all the workers are initialized and we can grab the pids.

src/users/amoc_users_sup.erl
```diff
@@ -1,60 +1,167 @@
...
+    UserSups = lists:reverse(supervisor:which_children(?MODULE)),
+    UserSupPids = [ Pid || {_, Pid, _, _} <- UserSups ],
+    UserSupPidsTuple = erlang:list_to_tuple(UserSupPids),
+    NumOfSupervisors = tuple_size(UserSupPidsTuple),
+    Atomics = atomics:new(1 + NumOfSupervisors, [{signed, false}]),
+    Storage = #storage{user_count = Atomics, sups = UserSupPidsTuple},
+    persistent_term:put(?MODULE, Storage).
+
+-spec clean_storage() -> any().
+clean_storage() ->
```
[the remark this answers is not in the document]
DenysGonchar: we don't really care about cleaning, because the value must be reset on init anyway.

src/users/amoc_users_sup.erl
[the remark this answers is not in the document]
DenysGonchar: right, the index... haven't noticed that

src/users/amoc_users_sup.erl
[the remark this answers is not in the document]
NelsonVides: Aaaaah that's what you mean, ok, thought you were referring to the `init` callback. Cool 👌🏽

src/amoc_controller.erl
```diff
@@ -182,12 +193,17 @@ handle_cast(_Msg, State) ->
...
-handle_info({'DOWN', _, process, Pid, _}, State) ->
-    NewSate = handle_stop_user(Pid, State),
+handle_info(start_all_users, State) ->
+    NewSate = handle_start_all_users(State),
     {noreply, NewSate};
 handle_info(_Msg, State) ->
     {noreply, State}.
 
+%% @private
+-spec terminate(term(), state()) -> any().
```
DenysGonchar, line 203: terminate is an optional callback, it can be removed

test/controller_SUITE.erl
```diff
@@ -54,6 +69,47 @@ end_per_testcase(_TestCase, Config) ->
...
+              begin
+                  SupervisorWithCounts = [ {dummy_pid(), Count} || Count <- Counts ],
+                  {Count, Assignments} = amoc_users_sup:distribute(Total, SupervisorWithCounts),
+                  Assinged = lists:sum([ N || {_, N} <- Assignments ]),
+                  Count =:= Total orelse Count =:= Assinged
+              end),
+    run_prop(?FUNCTION_NAME, Prop, 100, 1).
+
+% • The total count of the users is the exact sum of the values in the KV list
+total_count_is_the_exact_sum_of_the_values_in_the_input(_) ->
```
DenysGonchar, line 103: in the output

test/controller_SUITE.erl
```diff
@@ -1,18 +1,29 @@
...
 
 all() ->
     [
+     {group, distribute},
      {group, all_tests}
     ].
 
 groups() ->
     [
+     {distribute, [parallel], distribute()},
```
DenysGonchar, line 15: please add one more test:
* The number of items in the resulting KV list is equal to the number of non-zero items in the input KV list or requested amount (equal to the lower number)

test/controller_SUITE.erl
```diff
@@ -54,6 +69,47 @@ end_per_testcase(_TestCase, Config) ->
...
+              end),
+    run_prop(?FUNCTION_NAME, Prop, 100, 1).
+
+total_count_is_equal_to_the_requested_amount_or_the_sum_of_values_in_the_input(_) ->
+    Prop = ?FORALL({Total, Counts}, {pos_integer(), sups_with_counts()},
+              begin
+                  SupervisorWithCounts = [ {dummy_pid(), Count} || Count <- Counts ],
+                  {Count, Assignments} = amoc_users_sup:distribute(Total, SupervisorWithCounts),
+                  Assinged = lists:sum([ N || {_, N} <- Assignments ]),
+                  Count =:= Total orelse Count =:= Assinged
```
DenysGonchar, line 98: `Count =:= Assinged` is another property checked in the following TC, it should be like this:

```
Available = lists:sum([ N || {_, N} <- SupervisorWithCounts ]),
Count =:= min(Total, Available)
```

test/controller_SUITE.erl
```diff
@@ -54,6 +69,47 @@ end_per_testcase(_TestCase, Config) ->
...
+    Prop = ?FORALL({Total, Counts}, {pos_integer(), sups_with_counts()},
+              begin
+                  SupervisorWithCounts = [ {dummy_pid(), Count} || Count <- Counts ],
+                  {_, Assignments} = amoc_users_sup:distribute(Total, SupervisorWithCounts),
+                  Data = maps:from_list(SupervisorWithCounts),
+                  lists:all(fun({Pid, N}) -> N =< maps:get(Pid, Data) end, Assignments)
+              end),
+    run_prop(?FUNCTION_NAME, Prop, 100, 1).
+
+total_count_is_always_less_than_or_equal_the_requested_amount(_) ->
```
DenysGonchar, line 83: we can get rid of this TC if we check `Count =:= min(Total, Available)` in the following TC.

src/users/amoc_users_worker_sup.erl
```diff
@@ -0,0 +1,174 @@
...
+
+-spec get_all_children(pid()) -> [{pid(), amoc_scenario:user_id()}].
+get_all_children(Sup) ->
+    gen_server:call(Sup, get_all_children, infinity).
+
+%% @private
+-spec init(non_neg_integer()) -> {ok, term()}.
+init(N) ->
+    process_flag(trap_exit, true),
+    Tid = ets:new(?MODULE,  [ordered_set, protected]),
```
DenysGonchar, line 70: please make it named `list_to_atom(atom_to_list(__MODULE__) ++ integer_to_list(tid))`
so it's easy to identify the table.

src/users/amoc_users_sup.erl
```diff
@@ -1,60 +1,204 @@
...
+          sups_count :: pos_integer()
+         }).
+
+%% Supervisor
+
+%% @private
+-spec start_link() -> supervisor:startlink_ret().
+start_link() ->
+    Ret = supervisor:start_link({local, ?MODULE}, ?MODULE, no_args),
+    UserSups = lists:reverse(supervisor:which_children(?MODULE)),
```
DenysGonchar, line 43: please rewrite it in this way:
```
UserSups = supervisor:which_children(?MODULE),
IndexedSupsUnsorted = [ {Pid, N} || {{amoc_users_worker_sup, N}, Pid, _, _} <- UserSups],
IndexedSups = lists:keysort(2, IndexedSupsUnsorted),
UserSupPidsTuple = list_to_tuple([ Pid || {Pid, _} <- IndexedSups ]),
SupCount = tuple_size(UserSupPidsTuple),
Atomics = atomics:new(1 + SupCount, [{signed, false}]),
Storage = #storage{user_count = Atomics, sups = UserSupPidsTuple,
                   sups_indexed = IndexedSups, sups_count = SupCount},
persistent_term:put(?MODULE, Storage),
Ret.```

DenysGonchar (approved):
thanks for the huge amount of work, it looks really greate :)

PR #177, opened 2023-12-29
Add telemetry event for a coordinator trigger

NelsonVides:
Indicates an action is about to be triggered, either by enough users in the group or by timeout.


jacekwegr (approved):
Looks good, thank you for the changes!

PR #178, opened 2023-12-31
Supervision trees

NelsonVides:
This complex PR has a very important desire: introspection.

I want to be able to visualise, using for example observer or Livebook's supervision trees Explorer, the architecture of my load test. All coordinators and throttle processes are always supervised, named, and can easily be found.

It also serves a very small purpose of performance, as the coordinator for example is now not a single gen_event with a list of handlers to run sequentially, but, a pool of servers each one with its own handler, that can run in parallel.


test/throttle_SUITE.erl
```diff
@@ -0,0 +1,204 @@
...
+    %% send_and_wait passes fine
+    ?assertMatch(ok, amoc_throttle:send_and_wait(?FUNCTION_NAME, receive_this)),
+    %% pauses runs correctly
+    ?assertMatch(ok, amoc_throttle:pause(?FUNCTION_NAME)),
+    %% It is paused, so messages aren't received
+    amoc_throttle:send(?FUNCTION_NAME, receive_this),
+    ?assertMatch({error, not_received_yet}, receive_msg_in_timeout(receive_this, 1000)),
+    %% After resume the message is then received
+    ?assertMatch(ok, amoc_throttle:resume(?FUNCTION_NAME)),
+    ?assertMatch(ok, receive_msg_in_timeout(receive_this, 1000)).
```
DenysGonchar, line 156: please reduce this timeout.

DenysGonchar (changes requested):
This is nice PR, and it's cool to see that coverage has increased. but there are few things that must be addressed.

src/throttle/amoc_throttle_sup.erl
```diff
@@ -0,0 +1,57 @@
...
+%%
+%% The supervision tree is as follows
+%%
+%%                      amoc_sup
+%%                         |
+%%                 amoc_throttle_sup
+%%                 /       |        \
+%%                /        |         \
+%%               /         |          \
+%%           pooler    controller      pg
```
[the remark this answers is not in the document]
NelsonVides: This is not to decide which throttler belongs to which throttle, it is to simply supervise the workers, we never do `supervisor:which_children/1`. The thing is that we want a supervisor that will restart any worker in the pool, but pools aren't related between them. Having this multi-layer tree allows us to see all defined throttles and their worker sizes when drawing the supervision tree, if workers were added directly to the top supervisor then the tree would be a very dirty one.

Also pools can be calculated directly. The top supervisor is one_to_one, the pooler is the only one that is dynamic, each pool is again one_to_one. Having all workers directly under the top supervisor means that this sup needs to be one_to_one but then we add children to it dynamically, which is bad practice.

src/throttle/amoc_throttle_controller.erl
```diff
@@ -254,14 +256,14 @@ start_processes(Name, Rate, Interval, NoOfProcesses) ->
     raise_event(Name, init),
     RatePerMinute = rate_per_minute(Rate, Interval),
     report_rate(Name, RatePerMinute),
-    RealNoOfProcesses = min(Rate, NoOfProcesses),
+    RealNoOfProcesses = max(1, min(Rate, NoOfProcesses)),
```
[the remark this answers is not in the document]
NelsonVides: Rate can be zero, that means no throttling but simply having all workers execute all the time, this is allowed. I think that in this case we might just want to pass the number of workers as it was specified instead, without the `min`.

src/throttle/amoc_throttle_pool.erl
```diff
@@ -0,0 +1,57 @@
...
+    {ok, {supervisor:sup_flags(), [supervisor:child_spec()]}}.
+init({Name, Interval, Rate, NoOfProcesses}) ->
+    RatesPerProcess = calculate_rate_per_process(Rate, NoOfProcesses),
+    Tags = lists:seq(1, NoOfProcesses),
+    Children = [
+                #{id => {amoc_throttle_process, Name, N},
+                  start => {amoc_throttle_process, start_link, [Name, Interval, RatePerProcess]},
+                  type => worker,
+                  shutdown => timer:seconds(5),
+                  restart => permanent,
```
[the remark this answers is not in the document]
NelsonVides: Very good point, let's escalate this all the way up very loudly.

src/throttle/amoc_throttle_controller.erl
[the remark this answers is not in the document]
NelsonVides: Oh, no, I confused with interval set to zero, ok, correcting.

src/coordinator/amoc_coordinator_sup.erl
```diff
@@ -0,0 +1,67 @@
...
+            Other
+    end.
+
+-spec stop_coordinator(amoc_coordinator:name()) ->
+    ok | {error, term()}.
+stop_coordinator(Name) ->
+    case ets:lookup(?MODULE, Name) of
+        [{_, #{coordinator := Coordinator, timeout_worker := TimeoutWorker, workers := Workers}}] ->
+            [ amoc_coordinator_worker:stop(Worker) || Worker <- Workers ],
+            erlang:send(TimeoutWorker, terminate),
```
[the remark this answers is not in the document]
NelsonVides: This is from converting the previous code to a gen_server, sorry, fixing 😂

src/coordinator/amoc_coordinator_worker_sup.erl
```diff
@@ -0,0 +1,51 @@
...
+    Childs = [TimeoutChild | OrderedChilds],
+
+    SupFlags = #{strategy => one_for_one, intensity => 1, period => 5},
+    {ok, {SupFlags, Childs}}.
+
+%% Helpers
+timeout_child(Name, Timeout) ->
+    #{id => {amoc_coordinator_timeout, Name, Timeout},
+      start => {amoc_coordinator_timeout, start_link, [Name, Timeout]},
+      restart => permanent,
```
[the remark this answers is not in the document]
NelsonVides: Will change the supervision strategy to `intensity => 0`, that way any crash will immediately bubble up to the whole test, the scenario is buggy and shouldn't continue 👌🏽

src/coordinator/amoc_coordinator_worker_sup.erl
```diff
@@ -0,0 +1,51 @@
...
+      shutdown => timer:seconds(5),
+      type => worker,
+      modules => [amoc_coordinator_timeout]}.
+
+worker_spec(Item) ->
+    #{id => {amoc_coordinator_worker, Item},
+      start => {amoc_coordinator_worker, start_link, [Item]},
+      restart => permanent,
```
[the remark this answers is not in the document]
NelsonVides: I don't want to fetch the pids from the supervisor because then the supervisor is the bottleneck, fetching pids is a synchronous `gen_server:call/`. Anyway, I agree a failing coordinator is a bug so I'll rather bubble up the crash and stop the run.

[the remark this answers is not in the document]
NelsonVides: That actually would let me put the coordinator in a persistent term and skip the ets table, if we can not do any restart, pids don't need to change, and therefore persistent terms make sense. More performance 😄

test/throttle_SUITE.erl
```diff
@@ -0,0 +1,204 @@
...
+    ?assertMatch({ok, 99},
+                 amoc_throttle:change_rate(?FUNCTION_NAME, 100, ?DEFAULT_INTERVAL + 1)).
+
+change_rate_gradually(_) ->
+    ?assertMatch({error, {no_throttle_by_name, ?FUNCTION_NAME}},
+                 amoc_throttle:change_rate_gradually(
+                   ?FUNCTION_NAME, 100, 200, 1, 1, 1)),
+    ?assertMatch({ok, started},
+                 amoc_throttle:start(?FUNCTION_NAME, 100)),
+    ?assertMatch(ok,
```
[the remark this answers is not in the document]
NelsonVides: What do you mean by this one? 🤔

test/throttle_SUITE.erl
```diff
@@ -0,0 +1,204 @@
...
+    ?assertMatch({ok, started}, amoc_throttle:start(?FUNCTION_NAME, 100)).
+
+change_rate(_) ->
+    ?assertMatch({error, {no_throttle_by_name, ?FUNCTION_NAME}},
+                 amoc_throttle:change_rate(?FUNCTION_NAME, 100, ?DEFAULT_INTERVAL)),
+    ?assertMatch({ok, started},
+                 amoc_throttle:start(?FUNCTION_NAME, 100)),
+    ?assertMatch(ok,
+                 amoc_throttle:change_rate(?FUNCTION_NAME, 100, ?DEFAULT_INTERVAL)),
+    ?assertMatch({ok, 99},
```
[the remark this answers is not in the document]
NelsonVides: WTHF 🙄

src/throttle/amoc_throttle_controller.erl
```diff
@@ -153,28 +149,40 @@ handle_call({resume, Name}, _From, State) ->
...
 handle_call({change_rate, Name, Rate, Interval}, _From, State) ->
-    Info = maps:get(Name, State),
-    case maybe_change_rate(Name, Rate, Interval, Info) of
-        {ok, Rate} -> UpdatedInfo = Info#throttle_info{rate = Rate, interval = Interval},
-            {reply, ok, State#{Name => UpdatedInfo}};
-        Error -> {reply, Error, State}
+    case State of
+        #{Name := Info} ->
+            case maybe_change_rate(Name, Rate, Interval, Info) of
+                {ok, Rate} ->
```
NelsonVides, line 155: Right, this line is matching on it giving the provided value, if there's any diff this one doesn't match and the call will return that other rate that had been calculated.

test/throttle_SUITE.erl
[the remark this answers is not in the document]
NelsonVides: Explained here: https://github.com/esl/amoc/pull/178#discussion_r1451566294

src/throttle/amoc_throttle.erl
```diff
@@ -141,9 +141,13 @@ send(Name, Msg) ->
 %% Can be used to halt execution if we want a process to be idle when waiting for rate increase or other processes finishing their tasks.
 -spec send_and_wait(name(), any()) -> ok | {error, any()}.
 send_and_wait(Name, Msg) ->
-    send(Name, Msg),
-    receive
-        Msg -> ok
+    case send(Name, Msg) of
```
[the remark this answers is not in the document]
NelsonVides: Right, here we're going to have three processes waiting for each other's message, and if any dies the other two will lock in receive forever, so we need to set up careful monitoring. I'm already thinking `gen:call`, but even if doing the monitoring manually, I'd rather fix this "potential bug" in a separate PR with proper tests for it and separate it from this one as it differs in scope, size and intend. Will work on it immediately after this current one is merged 😉

src/coordinator/amoc_coordinator_worker_sup.erl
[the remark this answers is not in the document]
DenysGonchar: yes, coordinators should be started before spawning of the user processes, so persistent terms should be fine

src/throttle/amoc_throttle.erl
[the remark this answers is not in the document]
DenysGonchar: ok, let's put in a separate PR.

test/throttle_SUITE.erl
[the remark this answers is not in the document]
DenysGonchar: if we request to change the rate gradually within 10 minutes but then request another rate change during that 10 minutes, this second attempt should be discarded

src/throttle/amoc_throttle_controller.erl
DenysGonchar: damn, so `{ok, _}` is patternpamtched as Error clause...
this line should be changed to:
`{ok, NewRate} -> UpdatedInfo = Info#throttle_info{rate = NewRate, interval = Interval},`

NelsonVides: I don't think we need to store here the rate as returned by `maybe_change_rate`, that one is the rate given to each worker, which is `Rate div NoOfProcesses`. As it is currently, `throttle_info` will contain the newly provided rate, I think current code is already correct, have a look at the latest commit :)

test/throttle_SUITE.erl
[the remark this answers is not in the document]
NelsonVides: Added to the test 👌🏽

test/amoc_SUITE.erl
```diff
@@ -0,0 +1,122 @@
...
+start_and_then_stop_cannot_rerun(_) ->
+    Ret = amoc_do(testing_scenario, 1),
+    ?assertEqual(ok, Ret),
+    test_helpers:wait_until_scenario_has_users(testing_scenario, 1, 1),
+    Status = amoc:stop(),
+    ?assertMatch(ok, Status),
+    Retry = amoc_do(testing_scenario, 1),
+    ?assertMatch({{error, {invalid_status, _}}, 1}, Retry).
+
+after_rest_can_run_again(_) ->
```
DenysGonchar, line 105: typo

NelsonVides: fixed xD

test/controller_SUITE.erl
```diff
@@ -32,7 +32,9 @@ all_tests() ->
      stop_running_scenario_with_no_users_immediately_terminates,
      stop_running_scenario_with_users_stays_in_finished,
      stop_running_scenario_with_users_eventually_terminates,
-     scenario_with_state_and_crashing_in_terminate_run_fine
+     scenario_with_state_and_crashing_in_terminate_run_fine,
+     scenario_without_start_never_runs_users,
```
DenysGonchar, line 36: please suggest some better name

src/coordinator/amoc_coordinator.erl
```diff
@@ -0,0 +1,170 @@
...
+raise_telemetry_event(Name, Event) ->
+    TelemetryEvent = case Event of
+                         {coordinate, _} -> add;
+                         reset_coordinator -> reset;
+                         coordinator_timeout -> timeout
+                     end,
+    amoc_telemetry:execute([coordinator, TelemetryEvent], #{count => 1}, #{name => Name}).
+
+-spec order_plan([normalized_coordination_item()]) -> [normalized_coordination_item()].
+order_plan(Items) ->
```
DenysGonchar, line 140: please add a test for this function.
I must say that the version with `lists:partition/2` was more verbose.

+ The main idea of this function is that it defines an order of execution, the rest code must ensure that the order is kept and not broken. all other comments are irrelevant here. internal implementation can change, but execution order must remain as defined here. also, with the latest changes we no longer rely on the childrens' order.

test/telemetry_helpers.erl
```diff
@@ -0,0 +1,27 @@
...
+
+start(TelemetryEvents) ->
+    meck:new(?HANDLER, [non_strict, no_link]),
+    meck:expect(?HANDLER, handler, ['_', '_', '_', '_'], ok),
+    application:start(telemetry),
+    TelemetryHandler = fun ?HANDLER:handler/4,
+    telemetry:attach_many(?HANDLER, TelemetryEvents, TelemetryHandler, ?CONFIG).
+
+stop() ->
+    meck:reset(?HANDLER).
```
DenysGonchar, line 17: this is quite a confusing name for the function. there should be a separate stop function and a separate reset function.

test/throttle_SUITE.erl
```diff
@@ -0,0 +1,214 @@
...
+init_per_suite(Config) ->
+    application:ensure_all_started(amoc),
+    amoc_cluster:set_master_node(node()),
+    TelemetryEvents = [[amoc, throttle, Event] || Event <- [init, rate, request, execute, process]],
+    telemetry_helpers:start(TelemetryEvents),
+    Config.
+
+end_per_suite(_) ->
+    application:stop(amoc),
+    telemetry_helpers:stop(),
```
DenysGonchar, line 42: currently `telemetry_helpers:stop()` doesn't unload meck module, also we want to reset meck module on init_per_testcase to ensure that test cases do not affect each other.

test/amoc_coordinator_SUITE.erl
```diff
@@ -3,40 +3,41 @@
...
+end_per_suite(Config) ->
     application:stop(telemetry),
+    Sup = ?config(sup, Config),
+    Sup ! terminate,
     meck:unload().
 
 init_per_testcase(_, Config) ->
     meck:reset(?MOCK_MOD),
-    meck:reset(?TELEMETRY_HANDLER),
+    telemetry_helpers:stop(),
```
DenysGonchar, line 40: must be telemetry_helpers:reset()

test/amoc_coordinator_SUITE.erl
```diff
@@ -3,40 +3,41 @@
...
+    TelemetryEvents = [[amoc, coordinator, Event] || Event <- [start, stop, timeout, reset, add]],
+    telemetry_helpers:start(TelemetryEvents),
+    Pid = spawn(fun() -> amoc_coordinator_sup:start_link(), receive terminate -> ok end end),
+    [{sup, Pid} | Config].
 
-end_per_suite(_Config) ->
+end_per_suite(Config) ->
     application:stop(telemetry),
+    Sup = ?config(sup, Config),
+    Sup ! terminate,
```
DenysGonchar, line 35: that would be nice to call telemetry_helpers:stop() here, even if we do meck:unload() later

test/throttle_SUITE.erl
```diff
@@ -0,0 +1,214 @@
...
+assert_telemetry_event(Name, Count, Throttle, Rate, Interval) ->
+    TelemetryEvents = telemetry_helpers:get_calls([amoc, throttle]),
+    LowRateEvent = fun({EventName, Measurements, Metadata}) ->
+                           Name =:= EventName andalso
+                           1 =:= maps:get(Count, Measurements, 1) andalso
+                           Throttle =:= maps:get(name, Metadata, undefined) andalso
+                           Rate =:= maps:get(rate, Metadata, undefined) andalso
+                           Interval =:= maps:get(interval, Metadata, undefined)
+                   end,
+    lists:any(LowRateEvent, TelemetryEvents).
```
DenysGonchar, line 176: asserting is missing

test/throttle_SUITE.erl
```diff
@@ -0,0 +1,214 @@
...
+                   delay_between_executions := 600},
+                 State).
+
+
+%% Helpers
+assert_telemetry_event(Name, Count, Throttle, Rate, Interval) ->
+    TelemetryEvents = telemetry_helpers:get_calls([amoc, throttle]),
+    LowRateEvent = fun({EventName, Measurements, Metadata}) ->
+                           Name =:= EventName andalso
+                           1 =:= maps:get(Count, Measurements, 1) andalso
```
DenysGonchar, line 171: are we checking here equality to the default value?

test/throttle_SUITE.erl
```diff
@@ -0,0 +1,214 @@
...
+    ?assertMatch(#{name := ?FUNCTION_NAME,
+                   interval := 60000,
+                   delay_between_executions := 600},
+                 State).
+
+
+%% Helpers
+assert_telemetry_event(Name, Count, Throttle, Rate, Interval) ->
+    TelemetryEvents = telemetry_helpers:get_calls([amoc, throttle]),
+    LowRateEvent = fun({EventName, Measurements, Metadata}) ->
```
DenysGonchar, line 169: please rename into IsLowRateEventFn

test/throttle_SUITE.erl
```diff
@@ -0,0 +1,214 @@
...
+    ?assertMatch({error, invalid_throttle}, amoc_throttle:start(?FUNCTION_NAME, 0, 100, 1)).
+
+low_rate_gets_remapped(_) ->
+    ?assertMatch({ok, started}, amoc_throttle:start(?FUNCTION_NAME, 2, 100, 1)),
+    State = get_state_of_one_process(?FUNCTION_NAME),
+    ?assertMatch(#{name := ?FUNCTION_NAME,
+                   interval := 100,
+                   delay_between_executions := 50},
+                 State),
+    assert_telemetry_event([amoc, throttle, process], error, ?FUNCTION_NAME, 2, 100).
```
DenysGonchar, line 77: `'error'` atom is supplied as a `Count` parameter, this doesn't seem correct.

test/throttle_SUITE.erl
```diff
@@ -0,0 +1,214 @@
...
+
+low_interval_get_remapped(_) ->
+    ?assertMatch({ok, started}, amoc_throttle:start(?FUNCTION_NAME, 1, 1, 1)),
+    State = get_state_of_one_process(?FUNCTION_NAME),
+    ?assertMatch(#{name := ?FUNCTION_NAME,
+                   interval := 1,
+                   delay_between_executions := 10},
+                 State),
+    assert_telemetry_event([amoc, throttle, process], error, ?FUNCTION_NAME, 1, 1).
```
DenysGonchar, line 86: `'error'` atom is supplied as a `Count` parameter, this doesn't seem correct.

test/throttle_SUITE.erl
NelsonVides: We shouldn't need to. Test cases are running in parallel here, and also, when checking history we're filtering for the testcase name.

src/coordinator/amoc_coordinator.erl
NelsonVides: What we were doing before was respective the relative order given by the user between non-`all` elements and `all` elements, and simply moving the `all` items to the end. It is the same effect now, so this new code sets an equivalent order that it was already.

test/throttle_SUITE.erl
NelsonVides: We are checking that the key exists in the map, which wtf there's a `maps:is_key/2`, sorry xD

test/throttle_SUITE.erl
NelsonVides: It is: https://hexdocs.pm/amoc/telemetry.html#throttle-process-internals

test/throttle_SUITE.erl
NelsonVides: As above

src/coordinator/amoc_coordinator.erl
DenysGonchar: yeah, I know that it's the same, but lists:partition/2 version was easier to read.

test/throttle_SUITE.erl
DenysGonchar: I mean, that 'error' value for Count looks like a mistake at first glance. I would rather rename Count into a LogLevel.

NelsonVides: Renamed it to Measurement for consistency with the map.

src/coordinator/amoc_coordinator.erl
NelsonVides: All right, reverted then 👌🏽

DenysGonchar: please update the comment for order_plan/1 function:

```
The main idea of this function is that it defines an order of execution, the rest code must ensure that the order is kept and not broken. all other comments are irrelevant here. internal implementation can change, but execution order must remain as defined here. also, with the latest changes we no longer rely on the children's order.
```

I would suggest a comment like this:
```
This function defines the execution order of the events that must be processed synchronously (on reset/timeout/stop). we need to ensure that 'all' items are executed after 'non-all'.
Note that '{coordinate, _}' events are exceptional and their processing is done asynchronously.
The order of All plan items must be preserved as in the original plan, the same applies to NonAll items.
```

PR #180, opened 2024-01-17
Throttles/rework run and sending logic

NelsonVides:
An attempt to tackle the issue raised in https://github.com/esl/amoc/pull/178/files#r1451157854


src/throttle/amoc_throttle_controller.erl
```diff
@@ -98,6 +85,13 @@ stop(Name) ->
 telemetry_event(Name, Event) when Event =:= request; Event =:= execute ->
     raise_event(Name, Event).
 
+-spec raise_event_on_slave_node(name(), event()) -> ok.
+raise_event_on_slave_node(Name, Event) ->
```
DenysGonchar, line 89: Let's add a comment here that the purpuse of this function is to ensure that there are no event duplicates if we are running in a single node (non distributed) mode.

DenysGonchar (changes requested):

src/throttle/amoc_throttle_runner.erl
```diff
@@ -0,0 +1,61 @@
...
+        {'EXIT', RunnerPid, Reason} ->
+            exit({throttle_wait_died, RunnerPid, Reason});
+        '$scheduled' ->
+            ok
+    end;
+maybe_wait(_, _) ->
+    ok.
+
+-spec async_runner(amoc_throttle:name(), pid(), term()) -> no_return().
+async_runner(Name, Caller, Action) ->
```
DenysGonchar, line 45: asnyc runner must monitor trottler.

test/throttle_SUITE.erl
```diff
@@ -123,12 +128,50 @@ send_and_wait(_) ->
...
+    ?assertExit({throttle_wait_died, _, killed}, amoc_throttle:wait(?FUNCTION_NAME)).
+
+async_runner_dies_when_throttler_dies(_) ->
+    erlang:process_flag(trap_exit, true),
+    {links, OriginalLinks} = erlang:process_info(self(), links),
+    ?assertMatch({ok, started}, amoc_throttle:start(?FUNCTION_NAME, 1, 60000, 1)),
+    wait_until_one_throttle_worker(?FUNCTION_NAME),
+    amoc_throttle:send(?FUNCTION_NAME, receive_this),
+    wait_until_one_async_runner(self(), OriginalLinks),
+    kill_throttle_workers(?FUNCTION_NAME),
```
DenysGonchar, line 173: we can just stop the trottle with amoc_trottle:stop/1

NelsonVides: Oh, true :D

DenysGonchar (approved):
looks good to me.

PR #184, opened 2024-06-02
Throttle api

NelsonVides:
Making the API for throttle more friendly to use

### DIFF
--- a/guides/telemetry.md
+++ b/guides/telemetry.md
@@ -62,7 +62,7 @@ This event is raised only on the master node.
 
 ```erlang
 event_name: [amoc, throttle, rate]
-measurements: #{rate_per_minute := float()}
+measurements: #{rate := rate(), interval := interval()}
 metadata: #{monotonic_time := integer(), name := atom(), msg => binary()}
 ```
 
--- a/guides/telemetry.md
+++ b/guides/telemetry.md
@@ -62,7 +62,7 @@ This event is raised only on the master node.
 
 ```erlang
 event_name: [amoc, throttle, rate]
-measurements: #{rate := non_neg_integer()}
+measurements: #{rate_per_minute := float()}
 metadata: #{monotonic_time := integer(), name := atom(), msg => binary()}
 ```
 
--- a/guides/throttle.md
+++ b/guides/throttle.md
@@ -1,6 +1,6 @@
 ## API
 
-See `amoc_throttle`
+See `amoc_throttle`.
 
 ## Overview
 
@@ -9,9 +9,8 @@ It works in both local and distributed environments, allows for dynamic rate cha
 
 Amoc throttle allows to:
 
-- Setting the execution `Rate` per `Interval`
+- Setting the execution `Rate` per `Interval`, or inversely, the `Interarrival` time between actions.
 - Limiting the number of parallel executions when `Rate` is set to `infinity`
-- Setting the `Interarrival` time between actions.
 
 Each `Rate` is identified with a `Name`.
 The rate limiting mechanism allows responding to a request only when it does not exceed the given `Rate`.
--- a/guides/throttle.md
+++ b/guides/throttle.md
@@ -1,6 +1,6 @@
 ## API
 
-See `amoc_throttle`.
+See `m:amoc_throttle`.
 
 ## Overview
 
--- a/guides/throttle.md
+++ b/guides/throttle.md
@@ -10,13 +10,13 @@ It works in both local and distributed environments, allows for dynamic rate cha
 Amoc throttle allows to:
 
 - Setting the execution `Rate` per `Interval`, or inversely, the `Interarrival` time between actions.
-- Limiting the number of parallel executions when `Rate` is set to `infinity`
+- Limiting the number of parallel executions when `interval` is set to `0`.
 
-Each `Rate` is identified with a `Name`.
-The rate limiting mechanism allows responding to a request only when it does not exceed the given `Rate`.
-Amoc throttle makes sure that the given `Rate` per `Interval` is maintained on a constant level.
+Each throttle is identified with a `Name`.
+The rate limiting mechanism allows responding to a request only when it does not exceed the given throttle.
+Amoc throttle makes sure that the given throttle is maintained on a constant level.
 It prevents bursts of executions which could blurry the results, as they technically produce a desired rate in a given interval.
-Because of that, it may happen that the actual `Rate` would be slightly below the demanded rate. However, it will never be exceeded.
+Because of that, it may happen that the actual throttle rate would be slightly below the demanded rate. However, it will never be exceeded.
 
 ## Examples
 
@@ -46,18 +46,21 @@ user_loop(Id) ->
     user_loop(Id).
 ```
 Here a system should be under a continuous load of 100 messages per minute.
-Note that if we used something like `amoc_throttle:run(messages_rate, fun() -> send_message(Id) end)` instead of `amoc_throttle:send_and_wait/2` the system would be flooded with requests.
+Note that if we used something like `amoc_throttle:run(messages_rate, fun() -> send_message(Id) end)` instead of `amoc_throttle:wait/1` the system would be flooded with requests.
 
 A test may of course be much more complicated.
 For example it can have the load changing in time.
 A plan for that can be set for the whole test in `init/1`:
 ```erlang
 init() ->
-    %% init metrics
     amoc_throttle:start(messages_rate, 100),
     %% 9 steps of 100 increases in Rate, each lasting one minute
-    amoc_throttle:change_rate_gradually(messages_rate, 100, 1000, 60000, 60000, 9),
-    ok.
+    Gradual = #{from_rate => 100,
+                to_rate => 1000,
+                step_count => 9,
+                step_size => 100,
+                step_interval => timer:minutes(1)},
+    amoc_throttle:change_rate_gradually(messages_rate, Gradual).
 ```
 
 Normal Erlang messages can be used to schedule tasks for users by themselves or by some controller process.
@@ -101,7 +104,7 @@ For a more comprehensive example please refer to the `throttle_test` scenario, w
 - `amoc_throttle_controller.erl` - a gen_server which is responsible for reacting to requests, and managing `throttle_processes`.
 In a distributed environment an instance of `throttle_controller` runs on every node, and the one running on the master Amoc node stores the state for all nodes.
 - `amoc_throttle_process.erl` - gen_server module, implements the logic responsible for limiting the rate.
-For every `Name`, a `NoOfProcesses` are created, each responsible for keeping executions at a level proportional to their part of `Rate`.
+For every `Name`, a number of processes are created, each responsible for keeping executions at a level proportional to their part of the throttle.
 
 ### Distributed environment
 
@@ -116,12 +119,12 @@ Then a runner process is spawned on the same node.
 Its task will be to execute `Fun` asynchronously.
 A random throttle process which is assigned to the `Name` is asked for a permission for asynchronous runner to execute `Fun`.
 When the request reaches the master node, where throttle processes reside, the request metric on the master node is updated and the throttle process which got the request starts monitoring the asynchronous runner process.
-Then, depending on the system's load and the current rate of executions, the asynchronous runner is allowed to run the `Fun` or compelled to wait, because executing the function would exceed the calculated `Rate` in an `Interval`.
+Then, depending on the system's load and the current rate of executions, the asynchronous runner is allowed to run the `Fun` or compelled to wait, because executing the function would exceed the calculated throttle.
 When the rate finally allows it, the asynchronous runner gets the permission to run the function from the throttle process.
 Both processes increase the metrics which count executions, but for each the metric is assigned to their own node.
 Then the asynchronous runner tries to execute `Fun`.
 It may succeed or fail, either way it dies and an `'EXIT'` signal is sent to the throttle process.
-This way it knows that the execution of a task has ended, and can allow a different process to run its task connected to the same `Name` if the current `Rate` allows it.
+This way it knows that the execution of a task has ended, and can allow a different process to run its task connected to the same `Name` if the current throttle allows it.
 
 Below is a graph showing the communication between processes on different nodes described above.
 ![amoc_throttle_dist](assets/amoc_throttle_dist.svg)
--- a/guides/throttle.md
+++ b/guides/throttle.md
@@ -5,9 +5,14 @@ See `amoc_throttle`
 ## Overview
 
 Amoc throttle is a module that allows limiting the number of users' actions per given interval, no matter how many users there are in a test.
-It works in both local and distributed environments, allows for dynamic rate changes during a test and exposes metrics which show the number of requests and executions.
+It works in both local and distributed environments, allows for dynamic rate changes during a test and exposes telemetry events showing the number of requests and executions.
 
-Amoc throttle allows setting the execution `Rate` per `Interval` or limiting the number of parallel executions when `Interval` is set to `0`.
+Amoc throttle allows to:
+
+- Setting the execution `Rate` per `Interval`
+- Limiting the number of parallel executions when `Rate` is set to `infinity`
+- Setting the `Interarrival` time between actions.
+
 Each `Rate` is identified with a `Name`.
 The rate limiting mechanism allows responding to a request only when it does not exceed the given `Rate`.
 Amoc throttle makes sure that the given `Rate` per `Interval` is maintained on a constant level.
@@ -102,8 +107,8 @@ For every `Name`, a `NoOfProcesses` are created, each responsible for keeping ex
 ### Distributed environment
 
 #### Metrics
-In a distributed environment every Amoc node with a throttle started, exposes metrics showing the numbers of requests and executions.
-Those exposed by the master node show the sum of all metrics from all nodes.
+In a distributed environment every Amoc node with a throttle started, exposes telemetry events showing the numbers of requests and executions.
+Those exposed by the master node show the aggregate of all telemetry events from all nodes.
 This allows to quickly see the real rates across the whole system.
 
 #### Workflow
--- a/src/throttle/amoc_throttle.erl
+++ b/src/throttle/amoc_throttle.erl
@@ -86,17 +86,17 @@ start(Name, Rate) when is_integer(Rate) ->
 pause(Name) ->
     amoc_throttle_controller:pause(Name).
 
-%% @doc Unlocks executions for the given `Name' as if `Rate' was set to `infinity'.
--spec unlock(name()) -> ok | {error, any()}.
-unlock(Name) ->
-    amoc_throttle_controller:unlock(Name).
-
 %% @doc Resumes the executions for the given `Name', to their original configuration value.
 %%
 %% It is the counterpart to the `pause/1' API, resuming the execution of what that mechanism paused.
 -spec resume(name()) -> ok | {error, any()}.
 resume(Name) ->
     amoc_throttle_controller:resume(Name).
+
+%% @doc Unlocks executions for the given `Name' by setting `Rate' to `infinity'.
+-spec unlock(name()) -> ok | {error, any()}.
+unlock(Name) ->
+    change_rate(Name, #{rate => infinity, interval => 0}).
 
 %% @doc Sets the throttle `Config' for `Name' according to the given values.
 -spec change_rate(name(), t() | rate()) -> ok | {error, any()}.
--- a/src/throttle/amoc_throttle.erl
+++ b/src/throttle/amoc_throttle.erl
@@ -82,9 +82,11 @@ resume(Name) ->
 %%
 %% Can change whether Amoc throttle limits `Name' to parallel executions or to `Rate' per `Interval',
 %% according to the given `Interval' value.
--spec change_rate(name(), t()) -> ok | {error, any()}.
-change_rate(Name, Config) ->
-    amoc_throttle_controller:change_rate(Name, Config).
+-spec change_rate(name(), t() | rate()) -> ok | {error, any()}.
+change_rate(Name, #{} = Config) ->
+    amoc_throttle_controller:change_rate(Name, Config);
+change_rate(Name, Rate) when is_integer(Rate) ->
+    amoc_throttle_controller:change_rate(Name, #{rate => Rate}).
 
 %% @doc Allows to set a plan of gradual rate changes for a given `Name'.
 %%
--- a/src/throttle/amoc_throttle.erl
+++ b/src/throttle/amoc_throttle.erl
@@ -12,14 +12,16 @@
 %% Atom representing the name of the throttle.
 
 -type rate() :: infinity | non_neg_integer().
-%% Number of events per given `t:interval/0', or infinity for effectively unlocking all throttling.
-%% Note that a rate of zero means effectively pausing the throttle.
+%% Number of events per given `t:interval/0', or
+%% - infinity for effectively unlocking all throttling,
+%% - zero for effectively pausing the throttle.
 
 -type interarrival() :: infinity | non_neg_integer().
-%% Time in milliseconds between two events, or infinity for effectively pausing the throttle. Note
-%% that an interarrival of zero means effectively unlocking all throttling.
+%% Time in milliseconds between two events, or
+%% - infinity for effectively pausing the throttle,
+%% - zero for effectively unlocking all throttling.
 
--type interval() :: non_neg_integer().
+-type interval() :: pos_integer().
 %% In milliseconds, defaults to 60000 (one minute).
 
 -type throttle() :: #{rate := rate(), interval := interval()} |
--- a/src/throttle/amoc_throttle.erl
+++ b/src/throttle/amoc_throttle.erl
@@ -12,14 +12,22 @@
 %% Atom representing the name of the throttle.
 
 -type rate() :: infinity | non_neg_integer().
-%% Number of events per given `t:interval/0', or
-%% - infinity for effectively unlocking all throttling,
-%% - zero for effectively pausing the throttle.
+%% Number of events per given `t:interval/0'.
+%%
+%% It can also be:
+%% <ul>
+%%   <li>infinity for effectively unlocking all throttling,</li>
+%%   <li>zero for effectively pausing the throttle.</li>
+%% </ul>
 
 -type interarrival() :: infinity | non_neg_integer().
-%% Time in milliseconds between two events, or
-%% - infinity for effectively pausing the throttle,
-%% - zero for effectively unlocking all throttling.
+%% Time in milliseconds between two events.
+%%
+%% It can also be:
+%% <ul>
+%%   <li>infinity for effectively pausing the throttle,</li>
+%%   <li>zero for effectively unlocking all throttling.</li>
+%% </ul>
 
 -type interval() :: non_neg_integer().
 %% In milliseconds, defaults to 60000 (one minute).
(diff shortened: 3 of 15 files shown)


integration_test/helper.sh
```diff
@@ -18,6 +18,7 @@ function compile_file() {
 
 function contains_all() {
     local output="$(cat -)"
+    echo "Checking if \""$*"\" are contained in \"$output\""
```
DenysGonchar, line 21: to be removed.

NelsonVides: Not really, without that, if something fails in CI I never know what failed, for some reason unknown to me, all the `echo` commands below are muted, absolutely nothing is ever printed 🥲

src/throttle/amoc_throttle.erl
```diff
@@ -3,57 +3,63 @@
...
--define(NONNEG_INT(N), (is_integer(N) andalso N >= 0)).
--define(POS_INT(N), (is_integer(N) andalso N > 0)).
+-export([start/2, stop/1,
+         send/2, send/3, wait/1,
+         run/2, pause/1, resume/1, unlock/1,
+         change_rate/2, change_rate_gradually/2]).
 
 -type name() :: atom().
--type rate() :: pos_integer().
+%% Atom representing the name of the throttle.
```
DenysGonchar, line 12: it's so confusing to have comments below the code. please move comments above and add blank lines between definitions to make it readable. also, let's make it in edoc format.

NelsonVides: Nope, not possible to move the comments above and still generate the right documentation. It'd need to be migrated entirely to the newer documentation format.

DenysGonchar: thanks, it's much better with blank lines

src/throttle/amoc_throttle.erl
```diff
@@ -3,57 +3,68 @@
...
+%% Literal throttle configuration.
+
+-type gradual_rate_config() :: #{from_rate := non_neg_integer(),
+                                 to_rate := non_neg_integer(),
+                                 interval => interval(),
+                                 step_interval => pos_integer(),
+                                 step_size => pos_integer(),
+                                 step_count => pos_integer(),
+                                 duration => pos_integer()} |
+                               #{from_interarrival := interarrival(),
```
DenysGonchar, line 43: seems that `from_interarrival` cannot be 0 nor `infinity` in this case, the same is applicable to to_interarrival

src/throttle/amoc_throttle.erl
```diff
@@ -3,57 +3,68 @@
...
+
+-type throttle() :: #{rate := rate(), interval := interval()} |
+                    #{interarrival := interarrival()}.
+%% Throttle unit of measurement
+
+-type config() :: #{rate := rate(),
+                    interval => interval(),
+                    parallelism => non_neg_integer()}
+                | #{interarrival := interarrival(),
+                    parallelism => non_neg_integer()}.
```
DenysGonchar, line 33: what is the meaning of 0

src/throttle/amoc_throttle.erl
```diff
@@ -145,13 +159,6 @@ send(Name, Msg) ->
 send(Name, Pid, Msg) ->
     amoc_throttle_runner:throttle(Name, {Pid, Msg}).
 
-%% @doc Sends and receives the given message `Msg'.
-%%
-%% Deprecated in favour of `wait/1'
--spec send_and_wait(name(), any()) -> ok | {error, any()}.
```
DenysGonchar, line 151: I think that this interface is actively used by amoc_arsenal_xmpp. if it's so,  please add tickets for amoc_arsena/amoc_arsenal_xmpp

src/throttle/amoc_throttle.erl
```diff
@@ -67,12 +78,17 @@ pause(Name) ->
...
+    amoc_throttle_controller:unlock(Name).
+
+%% @doc Sets `Throttle' for `Name' according to the given values.
 %%
 %% Can change whether Amoc throttle limits `Name' to parallel executions or to `Rate' per `Interval',
 %% according to the given `Interval' value.
--spec change_rate(name(), rate(), interval()) -> ok | {error, any()}.
-change_rate(Name, Rate, Interval) ->
+-spec change_rate(name(), throttle()) -> ok | {error, any()}.
+change_rate(Name, #{rate := Rate, interval := Interval}) ->
```
DenysGonchar, line 91: don't we want to support `#{interarrival := interarrival()}` format here?

guides/throttle.md
```diff
@@ -5,9 +5,14 @@ See `amoc_throttle`
 ## Overview
 
 Amoc throttle is a module that allows limiting the number of users' actions per given interval, no matter how many users there are in a test.
-It works in both local and distributed environments, allows for dynamic rate changes during a test and exposes metrics which show the number of requests and executions.
+It works in both local and distributed environments, allows for dynamic rate changes during a test and exposes telemetry events showing the number of requests and executions.
+
+Amoc throttle allows to:
+
+- Setting the execution `Rate` per `Interval`
+- Limiting the number of parallel executions when `Rate` is set to `infinity`
```
DenysGonchar, line 13: `Rate` is supposed to be the number of parrallel executions if `Interval` is 0

guides/telemetry.md
```diff
@@ -62,7 +62,7 @@ This event is raised only on the master node.
 
 ```erlang
 event_name: [amoc, throttle, rate]
-measurements: #{rate := non_neg_integer()}
+measurements: #{rate_per_minute := float()}
```
DenysGonchar, line 65: recently we had to stick to rate_per_minute because of metrics reporting, but since we moved exometer support from amoc amoc_arsenal, do we still want to stick to "per minute" value?

guides/throttle.md
DenysGonchar: @NelsonVides, any comment here?

src/throttle/amoc_throttle.erl
```diff
@@ -3,57 +3,63 @@
...
+%% - infinity for effectively unlocking all throttling,
+%% - zero for effectively pausing the throttle.
+
+-type interarrival() :: infinity | non_neg_integer().
+%% Time in milliseconds between two events, or
+%% - infinity for effectively pausing the throttle,
+%% - zero for effectively unlocking all throttling.
+
+-type interval() :: pos_integer().
+%% In milliseconds, defaults to 60000 (one minute).
```
DenysGonchar, line 25: type should be non-negative integer, with zero for limiting parallel execution

src/throttle/amoc_throttle.erl
```diff
@@ -3,57 +3,63 @@
...
+                                 step_interval => pos_integer(),
+                                 step_size => pos_integer(),
+                                 step_count => pos_integer(),
+                                 duration => pos_integer()}.
+%% Configuration for a gradual throttle rate change
+%%
+%% `From' and `To' rates are required. `interval' defaults to 1s and `step_size' to 1 (or -1 if applies),
+%% that is, the throttle will be changed in increments of 1.
+%%
+%% All other values can be calculated from the provided.
```
DenysGonchar, line 49: are interval values given in ms? please add the comment.

NelsonVides: It is commented in line 24, where the interval type is defined. Only thing this isn't using such type, changing that now.

src/throttle/amoc_throttle.erl
```diff
@@ -3,57 +3,63 @@
...
 %% `Name' is needed to identify the rate as a single test can have different rates for different tasks.
-%% `Interval' is given in milliseconds and can be changed to a different value for convenience or higher granularity.
-%% It also accepts a special value of `0' which limits the number of parallel executions associated with `Name' to `Rate'.
--spec start(name(), rate(), interval(), pos_integer()) -> ok | {error, any()}.
-start(Name, Rate, Interval, NoOfProcesses)
-  when is_atom(Name), ?POS_INT(Rate), ?NONNEG_INT(Interval), ?POS_INT(NoOfProcesses) ->
-    amoc_throttle_controller:ensure_throttle_processes_started(Name, Rate, Interval, NoOfProcesses);
-start(_Name, _Rate, _Interval, _NoOfProcesses) ->
-    {error, invalid_throttle}.
+%% `Interval' is given in milliseconds, the default is one minute,
```
DenysGonchar, line 56: there is no Interval parameter for this interface; it is either the fully qualified config or the `Rate` (with the default Interval)

src/throttle/amoc_throttle.erl
```diff
@@ -3,57 +3,63 @@
...
+%% and can be changed to a different value for convenience or higher granularity.
+-spec start(name(), t() | rate()) -> {ok, started | already_started} | {error, any()}.
+start(Name, #{} = Config) ->
+    amoc_throttle_controller:ensure_throttle_processes_started(Name, Config);
+start(Name, Rate) ->
+    amoc_throttle_controller:ensure_throttle_processes_started(Name, #{rate => Rate}).
 
 %% @doc Pauses executions for the given `Name' as if `Rate' was set to `0'.
```
DenysGonchar, line 64: please add a comment that `resume/1` interface is the one to be used to undo pausing.

src/throttle/amoc_throttle.erl
```diff
@@ -62,18 +68,23 @@ start(_Name, _Rate, _Interval, _NoOfProcesses) ->
 pause(Name) ->
     amoc_throttle_controller:pause(Name).
 
+%% @doc Unlocks executions for the given `Name' as if `Rate' was set to `infinity'.
+-spec unlock(name()) -> ok | {error, any()}.
+unlock(Name) ->
+    amoc_throttle_controller:unlock(Name).
+
 %% @doc Resumes the executions for the given `Name', to their original `Rate' and `Interval' values.
```
DenysGonchar, line 76: please add a comment that this interface is a counterpart to `pause/1`

src/throttle/amoc_throttle.erl
NelsonVides: Huh, I changed these things a long time ago, let me refresh my thoughts...

The thing is that I don't think this is a good way to "limit parallel execution". What we were really doing was only limiting parallel "starting" of execution, but then there's the question of what if the executing action is long-timed? This wasn't waiting for any previous one to finish before starting a new one. See for example the difference between [`timer:apply_interval/4`](https://www.erlang.org/doc/apps/stdlib/timer.html#apply_interval/4) and [`timer:apply_repeatedly/4`](https://www.erlang.org/doc/apps/stdlib/timer.html#apply_repeatedly/4).

It also has the problem that it is an implementation detail we're leaking here. If you remember, we had this issue with not being able to set the rate too low or too high, because then it conflicts with the number of workers and makes the throttling wrong. When we gradually change a config, we can only change so much before the "parallelism" breaks the throttling calculations.

So here I'm taking a very complex change to actually hide entirely the number of workers and be able to enable or disable any of them as needed. And it is all tested.

If what we actually want is to limit parallel execution, then we'd need to consider whether waiting for actions to end or not as the `timer` module does, and that's an entirely different can of worms here.

src/throttle/amoc_throttle.erl
```diff
@@ -62,18 +68,23 @@ start(_Name, _Rate, _Interval, _NoOfProcesses) ->
...
 -spec resume(name()) -> ok | {error, any()}.
 resume(Name) ->
     amoc_throttle_controller:resume(Name).
 
-%% @doc Sets `Rate' and `Interval' for `Name' according to the given values.
+%% @doc Sets `Throttle' for `Name' according to the given values.
```
DenysGonchar, line 81: sets throttling `Config`, there's no parameter with the `Throttle` name

src/throttle/amoc_throttle.erl
NelsonVides: Ups 😅

src/throttle/amoc_throttle.erl
```diff
@@ -62,18 +68,23 @@ start(_Name, _Rate, _Interval, _NoOfProcesses) ->
...
 %% Can change whether Amoc throttle limits `Name' to parallel executions or to `Rate' per `Interval',
 %% according to the given `Interval' value.
--spec change_rate(name(), rate(), interval()) -> ok | {error, any()}.
-change_rate(Name, Rate, Interval) ->
-    amoc_throttle_controller:change_rate(Name, Rate, Interval).
+-spec change_rate(name(), t()) -> ok | {error, any()}.
+change_rate(Name, Config) ->
+    amoc_throttle_controller:change_rate(Name, Config).
 
 %% @doc Allows to set a plan of gradual rate changes for a given `Name'.
```
DenysGonchar, line 90: update the description for the `change_rate_gradually/1` interface.

src/throttle/amoc_throttle.erl
DenysGonchar: we were not starting any new action before any existing action finished. I'd like to keep this semantics.

NelsonVides: Same thing, what if, as the `timer` module does, I wanted to keep track of starting points and not full execution? And again, we were doing one or the other based on whether the given `fun` actually does stuff (for which the throttle blocks) or sends a message to allow someone else to do stuff (for which the throttle doesn't know).

This design is leaky, tremendously leaky. And the code where we could not have too high or too low rates because of the number of processes is not just more leaky but also ugly and limiting, it wouldn't allow me to run gradual changes to the edges of the config, which is something I really have a problem with.

And giving an interval of zero an ad-hoc meaning of "parallel executions" is even uglier, now we're mixing _three_ different ways to configure stuff (rate, interarrival, and parallelisation) with _two_ configuration keys. If throttle is a rate/interval computation, the intuitive thing to happen when interval gets lower and lower is for the throttle to get bigger and bigger, the fact that gets "capped" at a (paralellism) limit is an implementation detail we're not supposed to leak.

If what we want is a parallelism mechanism, I would actually give it a properly different API (more keys to be excluding with rate or interarrival as these are already excluding each other).

NelsonVides: Hmm... overloading a different API is all too ugly, but I'm going to experiment with adding a proper API for this. What I want the most is to get rid of the code that limits what rates can I set based on the parallelism given.

src/throttle/amoc_throttle_process.erl
```diff
@@ -143,46 +97,33 @@ handle_continue(maybe_run_fn, State) ->
...
-                {0, _, _} -> 0; %% limit only No of simultaneous executions
-                {_, I, _} when I < 10 ->
-                    Message = <<"too high rate, please increase NoOfProcesses">>,
-                    internal_error(Message, Name, Rate, Interval),
-                    10;
-                {_, DelayBetweenExecutions, 0} -> DelayBetweenExecutions;
-                {_, DelayBetweenExecutions, _} -> DelayBetweenExecutions + 1
-            end,
-    #state{interval = Interval, n = NewRate, max_n = NewRate, delay_between_executions = Delay}.
+initial_state(Name, infinity, 0) ->
```
DenysGonchar, line 113: there's no point in restricting delay value here, it can be like this:
initial_state(Name, infinity, _Delay) when is_integer(Delay) ->

src/throttle/amoc_throttle_process.erl
```diff
@@ -143,46 +97,33 @@ handle_continue(maybe_run_fn, State) ->
...
+    #state{name = Name, max_n = infinity, delay_between_executions = 0};
+initial_state(Name, MaxN, infinity) when is_integer(MaxN) ->
```
DenysGonchar, line 115: MaxN doesn't have to be an integer, pausing is possible even if MaxN is infinity.
you may want to reorder clauses to improve readability

src/throttle/amoc_throttle_process.erl
```diff
@@ -24,117 +16,79 @@
...
-    {error, no_throttle_process_registered} | {ok, [pid()]}.
-get_throttle_processes(Name) ->
-    case pg:get_members(?PG_SCOPE, Name) of
-        [] ->
-            {error, no_throttle_process_registered};
-        List ->
-            {ok, List}
-    end.
+-spec update(pid(), amoc_throttle:rate(), timeout()) -> ok.
+update(Pid, MaxN, Delay) ->
```
DenysGonchar, line 48: please add a comment that setting the delay to infinity results in effective pausing of the throttle process.

src/throttle/amoc_throttle_process.erl
```diff
@@ -143,46 +100,38 @@ handle_continue(maybe_run_fn, State) ->
...
-                {_, DelayBetweenExecutions, 0} -> DelayBetweenExecutions;
-                {_, DelayBetweenExecutions, _} -> DelayBetweenExecutions + 1
-            end,
-    #state{interval = Interval, n = NewRate, max_n = NewRate, delay_between_executions = Delay}.
-
-merge_state(#state{interval = I, delay_between_executions = D, n = N, max_n = MaxN},
-            #state{n = OldN, max_n = OldMaxN} = OldState) ->
+%% - If both `MaxN' and `Delay' are numbers, this will be the actual rate/interval.
+%%
+%% - If `MaxN' is a number and `Delay' is infinity, we mean to pause the process,
```
DenysGonchar, line 118: `MaxN` can be `infinity`, we don't care about it. it should be:
 - If `Delay' is infinity, we mean to pause the process

src/throttle/amoc_throttle_process.erl
```diff
@@ -143,46 +100,38 @@ handle_continue(maybe_run_fn, State) ->
...
+%%      see how at `maybe_start_timer/1' a delay of infinity will set `can_run_fn = false'.
+%%
+%% - If `MaxN' is infinity and `Delay' is zero, we mean no limits to throttling,
```
DenysGonchar, line 121: we don't care if `Delay` is zero, we check only if it's not equal to `infinity`:
- If `MaxN' is infinity and `Delay' is a number

src/throttle/amoc_throttle.erl
DenysGonchar: also please remove references to non-existing  variables `From` and `To`

src/throttle/amoc_throttle.erl
```diff
@@ -1,133 +1,149 @@
...
 %%
-%% `Rate' will be changed from `FromRate' to `ToRate' in a series of consecutive steps.
-%% Note that `FromRate' does not need to be lower than `ToRate', rates can be changed downwards.
+%% The configuration will be changed in a series of consecutive steps.
+%% Rates can be changed upwards as well as downwards.
 %%
 %% The rate is calculated at each step in relation to the `RateInterval', which can also be `0'.
 %% There will be `NoOfSteps' steps, each taking `StepInterval' time in milliseconds.
 %%
```
DenysGonchar, line 103: please remove references to non-existing parameters (RateInterval,  NoOfSteps, StepInterval) and use corresponding atom keys in the configuration map.

guides/throttle.md
```diff
@@ -1,13 +1,17 @@
 ## API
 
-See `amoc_throttle`
+See `m:amoc_throttle`.
```
DenysGonchar, line 3: please update the whole document and remove references to non-existing parameters (Rate, Interval, Interarrival, etc.). use corresponding atom keys in the configuration map instead.

src/throttle/amoc_throttle_controller.erl
```diff
@@ -203,123 +247,291 @@ handle_info({change_plan, Name}, State) ->
...
+%% Helpers
+
+-spec pool_config(amoc_throttle:rate(), amoc_throttle:interval()) -> pool_config().
+pool_config(infinity, _) ->
+    Config = #{max_n => infinity, delay => 0, status => active, pid => undefined},
+    maps:from_keys(lists:seq(1, no_of_processes()), Config);
+pool_config(0, _) ->
+    Config = #{max_n => 0, delay => infinity, status => active, pid => undefined},
+    maps:from_keys(lists:seq(1, no_of_processes()), Config);
+pool_config(Rate, 0) ->
```
DenysGonchar, line 347: seems that this is a clause to be changed :)

src/throttle/amoc_throttle_pool.erl
```diff
@@ -5,52 +5,30 @@
...
-        amoc_throttle:rate(),
-        pos_integer()
-       ) -> ok | error.
-start_process_pool(Name, Interval, Rate, NoOfProcesses) ->
-    {ok, _} = supervisor:start_child(amoc_throttle_pooler, [Name, Interval, Rate, NoOfProcesses]),
-    ok.
+-spec start_link(amoc_throttle:name(), amoc_throttle_config:pool_config()) ->
+    supervisor:startlink_ret().
+start_link(Name, PoolConfig) ->
+    case supervisor:start_link(?MODULE, {Name, PoolConfig}) of
```
DenysGonchar, line 13: what is the point of keeping this `case`?

src/throttle/amoc_throttle_process.erl
```diff
@@ -195,6 +145,11 @@ maybe_stop_timer(#state{tref = TRef}) ->
     {ok, cancel} = timer:cancel(TRef),
     consume_all_timer_ticks(delay_between_executions).
 
+timeout(#state{delay_between_executions = infinity}) ->
+    ?DEFAULT_MSG_TIMEOUT;
```
DenysGonchar, line 149: I think it should be set to `infinity` in this case.

src/throttle/amoc_throttle_config.erl
```diff
@@ -0,0 +1,188 @@
...
+    PoolConfig.
+
+-spec no_of_processes() -> non_neg_integer().
+no_of_processes() ->
+    3 * erlang:system_info(schedulers_online).
+
+-spec do_verify_gradual_config(amoc_throttle:gradual_rate_config()) -> gradual().
+do_verify_gradual_config(
+  #{from_interarrival := FromInterarrival, to_interarrival := ToInterarrival} = Config0)
+  when ?POS_INT(FromInterarrival), ?POS_INT(ToInterarrival) ->
```
DenysGonchar, line 123: when ?POS_INT(FromInterarrival), ?POS_INT(ToInterarrival), not is_map_key(interval, Config0) ->

guides/throttle.md
NelsonVides: Fixed in a newer version of the code.

src/throttle/amoc_throttle_config.erl
```diff
@@ -0,0 +1,188 @@
...
+-define(DEFAULT_INTERVAL, 60000). %% one minute
+-define(DEFAULT_STEP_SIZE, 1).
+
+-export([verify_config/1, verify_gradual_config/1, pool_config/2, process_pool_config/2]).
+-export_type([config/0, gradual/0, pool_config/0]).
+
+-type process_number() :: non_neg_integer().
+-type config() :: #{rate := amoc_throttle:rate(),
+                    interval := amoc_throttle:interval()}.
+-type gradual() :: #{from_rate := non_neg_integer(),
```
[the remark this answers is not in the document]
DenysGonchar: or maybe even:
• steps - list of {rate, interval} tuples
• step_interval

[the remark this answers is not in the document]
DenysGonchar: this will be the most versatile for the controller.

DenysGonchar (changes requested):

src/throttle/amoc_throttle.erl
NelsonVides: It is already updated 👌🏽

src/throttle/amoc_throttle_controller.erl
```diff
@@ -8,31 +8,33 @@
...
-    change_plan :: change_rate_plan() | undefined
+    active = true :: boolean(),
+    change_plan :: undefined | change_rate_plan()
 }).
 
 -record(change_rate_plan, {
     high_rate :: pos_integer(),
     no_of_steps :: non_neg_integer(),
+    step_size :: pos_integer(),
     timer :: timer:tref()}).
```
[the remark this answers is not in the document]
NelsonVides: Hmm... actually, that sounds really neat... 🤔

src/throttle/amoc_throttle_config.erl
```diff
@@ -0,0 +1,188 @@
...
+do_verify_gradual_config(
+  #{from_rate := FromRate, to_rate := ToRate,
+    step_interval := StepInterval, step_count := StepCount, step_size := StepSize} = Config) ->
+    RateInterval = maps:get(interval, Config, ?DEFAULT_INTERVAL),
+    check_step_size_with_from_to_rate(FromRate, ToRate, StepSize),
+    check_step_parameters((ToRate - FromRate) div StepCount, StepSize),
+    #{from_rate => FromRate, to_rate => ToRate, interval => RateInterval,
+      step_interval => StepInterval, step_count => StepCount, step_size => StepSize};
+do_verify_gradual_config(
+  #{from_rate := FromRate, to_rate := ToRate, duration := Duration}) ->
```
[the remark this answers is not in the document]
NelsonVides: Yeah, this function is getting very wild, I'll happily simplify it 👌🏽

src/throttle/amoc_throttle_config.erl
```diff
@@ -0,0 +1,157 @@
...
+    Fun1 = fun(N, Config) -> Config#{pid => maps:get(N, Workers)} end,
+    maps:map(Fun1, PoolConfig).
+
+-spec calculate_availability(integer(), integer(), pos_integer(), integer()) -> pool_config().
+calculate_availability(RatePerMinutePerProcess, DelayPerProcess, NoOfProcesses, Rem) ->
+    Fun = fun(N, {Acc, R}) ->
+                  case {RatePerMinutePerProcess < NoOfProcesses, R} of
+                      {true, 0} ->
+                          Config = #{max_n => RatePerMinutePerProcess,
+                                     delay => DelayPerProcess + 1,
```
DenysGonchar, line 89: why `Rem` is added to `delay` but not `max_n`

DenysGonchar: I really don't get this function

src/throttle/amoc_throttle_controller.erl
```diff
@@ -203,123 +205,131 @@ handle_info({change_plan, Name}, State) ->
...
-            {reply, {error, wrong_no_of_procs}, State}
-    end.
-
-run_cmd(Pid, stop) ->
-    amoc_throttle_process:stop(Pid);
-run_cmd(Pid, pause) ->
-    amoc_throttle_process:pause(Pid);
-run_cmd(Pid, resume) ->
-    amoc_throttle_process:resume(Pid).
+do_run_op(stop, Name, #throttle_info{pool_sup = PoolSup}, State) ->
```
DenysGonchar, line 280: why don't we update process group here?

src/throttle/amoc_throttle_controller.erl
```diff
@@ -203,123 +205,131 @@ handle_info({change_plan, Name}, State) ->
...
+          end,
+    maps:foreach(Fun, PoolConfig),
+    {reply, ok, State#{Name => Info#throttle_info{active = false}}};
+do_run_op(unlock, Name, #throttle_info{pool_config = PoolConfig} = Info, State) ->
+    Fun = fun(_, #{pid := Pid}) ->
+                  amoc_throttle_process:update(Pid, infinity, 0)
+          end,
+    maps:foreach(Fun, PoolConfig),
+    {reply, ok, State#{Name => Info#throttle_info{active = true}}};
+do_run_op(resume, Name, #throttle_info{pool_config = PoolConfig} = Info, State) ->
```
DenysGonchar, line 295: I would expect the resume operation to work only if `throttle_info{active = false}`

src/throttle/amoc_throttle_controller.erl
```diff
@@ -203,123 +205,131 @@ handle_info({change_plan, Name}, State) ->
...
+do_run_op(stop, Name, #throttle_info{pool_sup = PoolSup}, State) ->
+    ok = amoc_throttle_pooler:stop_pool(PoolSup),
+    {reply, ok, maps:remove(Name, State)};
+do_run_op(pause, Name, #throttle_info{pool_config = PoolConfig} = Info, State) ->
+    Fun = fun(_, #{pid := Pid}) ->
+                  amoc_throttle_process:update(Pid, 0, infinity)
+          end,
+    maps:foreach(Fun, PoolConfig),
+    {reply, ok, State#{Name => Info#throttle_info{active = false}}};
+do_run_op(unlock, Name, #throttle_info{pool_config = PoolConfig} = Info, State) ->
```
DenysGonchar, line 289: I don't like this unlock concept, also there's no trace in the state that unlocking has happened. 
from my perspective, there should be no `amoc_throttle_controller:unlock/1` API. if you want to have `amoc_throttle:unlock/1` API, then it should just call `amoc_throttle_controller:change_rate/2`

src/throttle/amoc_throttle_controller.erl
NelsonVides: This will stop the running processes and `pg` will get a monitor down message and remove them automatically.

test/throttle_SUITE.erl
```diff
@@ -64,31 +71,61 @@ start(_) ->
...
+    ?assertMatch(#{interval := 60000, rate := 0}, State).
+
+start_rate_zero(_) ->
+    %% Starts successfully
+    Description = #{rate => 0},
+    ?assertMatch({ok, started}, amoc_throttle:start(?FUNCTION_NAME, Description)),
+    State = get_throttle_info(?FUNCTION_NAME),
+    ?assertMatch(#{interval := 60000, rate := 0}, State).
+
+start_rate_infinity(_) ->
```
DenysGonchar, line 109: start_interval_zero and start_interval_infinity test cases are missing

test/throttle_SUITE.erl
```diff
@@ -102,107 +139,242 @@ start_and_stop(_) ->
...
-                 amoc_throttle:change_rate(?FUNCTION_NAME, 100, ?DEFAULT_INTERVAL)),
+                 amoc_throttle:change_rate(?FUNCTION_NAME, #{rate => 100})),
+    ?assertMatch({ok, started}, amoc_throttle:start(?FUNCTION_NAME, 100)),
+    E1 = #{rate => 100, interval => ?DEFAULT_INTERVAL},
+    ?assertMatch(ok, amoc_throttle:change_rate(?FUNCTION_NAME, E1)),
+    E2 = #{rate => 100, interval => ?DEFAULT_INTERVAL + 1},
+    ?assertMatch(ok, amoc_throttle:change_rate(?FUNCTION_NAME, E2)),
+    E3 = #{rate => 100, interval => ?DEFAULT_INTERVAL + 2},
+    ?assertMatch(ok, amoc_throttle:change_rate(?FUNCTION_NAME, E3)),
+    ?assertMatch(ok, amoc_throttle:change_rate(?FUNCTION_NAME, 200)).
```
DenysGonchar, line 150: hot about testing 0 and infinity rate and interval. what about interarrival?

src/throttle/amoc_throttle_config.erl
NelsonVides: Let me try to add some comments with explanations.

test/throttle_SUITE.erl
DenysGonchar: right, that a valid value only for delay at `amoc_throttle_process`

src/throttle/amoc_throttle_config.erl
NelsonVides: Ok, I've added a property-based test to check that the aggregated delays given to all processes don't differ from the actually requested rate by more than 1%, and I cannot get it to pass 😱 

Please have a look at `throttle_SUITE:pool_config_property_tests/1` @DenysGonchar

DenysGonchar (approved):

PR #185, opened 2024-06-14
Cosmetics changes to improve readability, types, and upgrades

NelsonVides:


src/amoc_scenario.erl
```diff
@@ -73,7 +73,7 @@ terminate(Scenario, State) ->
 %% if scenario module exports both functions, `Scenario:start/2' is used.
 %%
 %% Runs on the user process and spans a `[amoc, scenario, user, _]' telemetry event.
--spec start(amoc:scenario(), user_id(), state()) -> any().
+-spec start(amoc:scenario(), user_id(), state()) -> term().
```
DenysGonchar, line 76: `any()` highlights that returned value is ignored.

integration_test/extra_code_paths/path1/dummy_helper.erl
```diff
@@ -11,65 +16,73 @@ test_amoc_dist() ->
...
-        N1 = Max1,
-        Ids1 = lists:seq(1, N1),
+        ?assertEqual(disabled, get_status(Master)),
+        [ ?assertMatch({running, #{scenario := dummy_scenario}}, get_status(Node)) || Node <- Slaves],
+        %% check user ids, users have all been started at the first two nodes
+        {N1, Max1, Nodes1, Ids1, Users1} = get_users_info(Slaves),
+        ?assert(N1 > 0),
+        ?assertEqual(N1, Max1, ?comment(Users1)),
+        ?assertEqual(Ids1, lists:seq(1, N1), ?comment(Users1)),
         [AddedNode] = Slaves -- Nodes1,
```
DenysGonchar, line 26: why not assertEqual?

NelsonVides: Because then the value `AddedNode` is not captured and it is used a little bit below. Could do so nevertheless, only it would require two lines of code instead of one :)

DenysGonchar: right

src/amoc_sup.erl
```diff
@@ -11,14 +11,14 @@
 -export([init/1]).
 
 %% Helper macro for declaring children of supervisor
--define(WORKER(I, Type), {I, {I, start_link, []}, permanent, 5000, Type, [I]}).
--define(SUP(I, Type), {I, {I, start_link, []}, permanent, infinity, Type, [I]}).
+-define(WORKER(I), {I, {I, start_link, []}, permanent, 5000, worker, [I]}).
+-define(SUP(I), {I, {I, start_link, []}, permanent, infinity, supervisor, [I]}).
```
DenysGonchar, line 15: yeah, this was a bit stupid :)

src/users/amoc_users_sup.erl
```diff
@@ -139,14 +139,14 @@ terminate_all_children() ->
 -spec get_sup_for_user_id(amoc_scenario:user_id()) -> pid().
 get_sup_for_user_id(Id) ->
     #storage{sups = Supervisors, sups_count = SupCount} = persistent_term:get(?MODULE),
-    Index = Id rem SupCount + 1,
+    Index = erlang:phash2(Id, SupCount) + 1,
```
DenysGonchar, line 142: I'm not sure if phash is making a better distribution then a simple rem.

DenysGonchar (approved):
looks good

PR #186, opened 2024-06-14
Interarrival as throttle

NelsonVides:


guides/configuration.md
```diff
@@ -9,11 +9,9 @@ Amoc supports the following generic configuration parameters:
     * default value - empty list (`[]`)
     * example: `AMOC_NODES="['amoc@amoc-1', 'amoc@amoc-2']"`
 
-
-* `interarrival` - a delay (in ms, for each node in the cluster independently) between creating the processes
-  for two consecutive users:
-    * default value - 50 ms.
-    * example: `AMOC_INTERARRIVAL="50"`
+* `user_rate` - a rate (in ms, implemented as a `t:amoc_throttle:rate/0`) of user processes per minute:
```
DenysGonchar, line 12: remove `in ms`

src/amoc.erl
```diff
@@ -25,7 +25,7 @@ do(Scenario, Count, Settings) ->
             case {amoc_controller:start_scenario(Scenario, Settings), Count} of
                 {ok, 0} -> ok;
                 {ok, Count} -> amoc_controller:add_users(1, Count);
-                Error -> Error
+                {Error, _} -> Error
```
DenysGonchar, line 28: why? according to the spec it was correct before the change:
```
-spec start_scenario(amoc:scenario(), amoc_config:settings()) ->
ok | {error, term()}.
```

src/amoc_controller.erl
```diff
@@ -8,40 +8,46 @@
...
+-define(DEFAULT_USER_RATE, 1200).
 
--required_variable(#{name => interarrival, default_value => 50,
-                     verification => {?MODULE, non_neg_integer, 1},
-                     description => "a delay between creating the processes for two "
-                                    "consecutive users (ms, def: 50ms)",
-                     update => {?MODULE, maybe_update_interarrival_timer, 2}}).
+-required_variable(#{name => user_rate, default_value => ?DEFAULT_USER_RATE,
+                     verification => {?MODULE, verify_user_rate, 1},
+                     description => "Throttle rate for the Scenario:start/1,2 callback (def: 50ms)",
```
DenysGonchar, line 15: remove `(def: 50ms)`

src/amoc_controller.erl
```diff
@@ -122,14 +128,16 @@ disable() ->
     gen_server:call(?SERVER, disable).
 
 %% @private
--spec non_neg_integer(any()) -> boolean().
-non_neg_integer(Interarrival) ->
-    is_integer(Interarrival) andalso Interarrival >= 0.
+-spec verify_user_rate(any()) -> boolean().
+verify_user_rate(UserRate) ->
```
DenysGonchar, line 132: please use braces to make expression more readable

src/throttle/amoc_throttle_controller.erl
```diff
@@ -260,16 +260,17 @@ continue_plan(Name, State, Info, #change_rate_plan{rates = [Rate]} = Plan) ->
     TRef = Plan#change_rate_plan.timer,
     Info1 = do_change_rate(Name, Rate, Interval, Info),
     {ok, cancel} = timer:cancel(TRef),
-    consume_all_timer_ticks({change_plan, Name}),
+    consume_all_messages({change_plan, Name}),
```
DenysGonchar, line 263: revert this change, "consume_all_timer_ticks" is more readable name for this function

src/throttle/amoc_throttle_process.erl
```diff
@@ -143,19 +143,13 @@ maybe_stop_timer(#state{tref = undefined}) ->
     ok;
 maybe_stop_timer(#state{tref = TRef}) ->
     {ok, cancel} = timer:cancel(TRef),
-    consume_all_timer_ticks(delay_between_executions).
+    amoc_throttle_controller:consume_all_messages(delay_between_executions).
```
DenysGonchar, line 146: revert this change, it introduces erronous impression that we do here something with amoc_throttle controller interface.

src/amoc.erl
NelsonVides: `Error` was `{{error, Reason}, Count}`, which is not the spec. Here we want to return the error that `amoc_controller:start_scenario/2` returned.

src/throttle/amoc_throttle_process.erl
NelsonVides: So where would you put it? The code was duplicated, wanted to put it in some common helper.

src/amoc.erl
DenysGonchar: right, sorry, dunno how I missed that :)

guides/configuration.md
```diff
@@ -9,11 +9,9 @@ Amoc supports the following generic configuration parameters:
     * default value - empty list (`[]`)
     * example: `AMOC_NODES="['amoc@amoc-1', 'amoc@amoc-2']"`
 
-
-* `interarrival` - a delay (in ms, for each node in the cluster independently) between creating the processes
-  for two consecutive users:
-    * default value - 50 ms.
-    * example: `AMOC_INTERARRIVAL="50"`
+* `user_rate` - a rate (implemented as a `t:amoc_throttle:rate/0`) of user processes per minute:
```
DenysGonchar, line 12: what is `t:amoc_throttle:rate/0`?

src/amoc_controller.erl
DenysGonchar:
```
(is_integer(UserRate) andalso (0 =< UserRate))
     orelse  (infinity =:= UserRate)
```

src/throttle/amoc_throttle_process.erl
DenysGonchar: if we don't have any suitable helper module for this, then I would rather want to duplicate code.

src/users/amoc_users_sup.erl
```diff
@@ -4,7 +4,7 @@
 %%
 %% It spawns a pool of workers as big as online schedulers. When starting a new user, as the user is
 %% identified by ID, a worker will be chosen for this user based on its ID
-%% (see get_sup_for_user_id/1).
+%% (see `gen_sup_from_userid/1').
```
DenysGonchar, line 7: why `gen_sup_from_userid/1`?  the old name seems to be more relevant.

src/users/amoc_users_sup.erl
```diff
@@ -93,24 +100,24 @@ decr_no_of_users(SupNum) when SupNum > 1 ->
...
 %% depending on the number of users.
--spec start_children(amoc:scenario(), [amoc_scenario:user_id()], any()) -> ok.
-start_children(Scenario, UserIds, ScenarioState) ->
-    State = persistent_term:get(?MODULE),
-    #storage{sups = Supervisors, sups_indexed = IndexedSups, sups_count = SupCount} = State,
-    Acc = maps:from_list([ {Sup, []} || {Sup, _} <- IndexedSups ]),
-    Assignments = assign_users_to_sups(SupCount, Supervisors, UserIds, Acc),
+-spec start_children(amoc:scenario(), {amoc_scenario:user_id(), amoc_scenario:user_id()}, any()) ->
+    ok.
+start_children(Scenario, {StartId, EndId}, ScenarioState) ->
```
DenysGonchar, line 118: let's convert `StartId`, `EndId` into normal function parameters, tuple is unneeded here.

src/users/amoc_users_sup.erl
```diff
@@ -76,14 +81,16 @@ count_no_of_users() ->
     #storage{user_count = Atomics} = persistent_term:get(?MODULE),
     atomics:get(Atomics, 1).
 
--spec incr_no_of_users(non_neg_integer()) -> any().
-incr_no_of_users(SupNum) when SupNum > 1 ->
+-spec handle_up_user(non_neg_integer(), pid(), amoc_scenario:user_id()) -> any().
+handle_up_user(SupNum, Pid, Id) when SupNum > 1 ->
+    ets:insert(?TABLE, {Pid, Id}),
```
DenysGonchar, line 86: can we just put `user_id` into the process dictionary of the user process? and add an interface at `amoc_user` to get `user_id` by `pid` (must be an rpc call for remote processes since process_info interface works only for the local processes)

src/users/amoc_users_sup.erl
```diff
@@ -34,14 +33,17 @@
...
+
 %% Supervisor
 
 %% @private
 -spec start_link() -> supervisor:startlink_ret().
 start_link() ->
     Ret = supervisor:start_link({local, ?MODULE}, ?MODULE, no_args),
     UserSups = supervisor:which_children(?MODULE),
-    IndexedSupsUnsorted = [ {Pid, N} || {{amoc_users_worker_sup, N}, Pid, _, _} <- UserSups],
+    IndexedSupsUnsorted = [ {Pid, N} || {{amoc_users_worker_sup, N}, Pid, _, _} <- UserSups,
```
DenysGonchar, line 45: could you remind me why it's unsorted?

DenysGonchar: and please add a comment at `indexes/0` that we start indexes from 2 to simplify user_count atomics management.

src/users/amoc_users_sup.erl
```diff
@@ -34,14 +33,17 @@
...
+                                        is_integer(N), is_pid(Pid)],
```
DenysGonchar, line 46: how is it possible that these filters could return false?

src/users/amoc_users_sup.erl
```diff
@@ -120,47 +127,53 @@ start_children(Scenario, UserIds, ScenarioState) ->
...
+               '$end_of_table' ->
+                   [];
+               {Objects, _} ->
+                   Objects
+           end,
+    stop_children_assignments(Users, Force),
+    length(Users).
 
 -spec get_all_children() -> [{pid(), amoc_scenario:user_id()}].
 get_all_children() ->
```
DenysGonchar, line 140: seems that this function is used for testing only, we can remove it and rework testing a bit.

test/controller_SUITE.erl
```diff
@@ -214,8 +215,9 @@ stop_running_scenario_with_users_eventually_terminates(_) ->
     WaitUntilValue = {finished, testing_scenario},
     wait_helper:wait_until(WaitUntilFun, WaitUntilValue).
 
-interarrival_equal_zero_starts_all_users_at_once(_) ->
-    Vars = [{interarrival, 0}, {testing_var1, def1} | test_helpers:other_vars_to_keep_quiet()],
+user_rate_equal_zero_starts_all_users_at_once(_) ->
+    Vars = [{user_rate, 0}, {testing_var1, def1}
```
DenysGonchar, line 219: it must be `infinity` user rate, not 0.

src/users/amoc_users_sup.erl
NelsonVides: I'm not sure to understand what does that give. This is called by the supervisor worker, not the user process 🤔

src/users/amoc_users_sup.erl
NelsonVides: That the first is an integer, not possible, but only we know that. That the second is not a pid can happen if a children is being restarted. Which is probably a catastrophe. So in all, this is kind of an artifact to make Eqwalizer happy that is is really strictly a list of {Pid,N} pairs.

src/users/amoc_users_sup.erl
NelsonVides: They'll be returned in reversed order from where they were started, so if they were declared as 1..12, this will return 12..1, so maybe even a reverse would do. But just to be safe I thought of a keysort.

src/users/amoc_users_sup.erl
NelsonVides: Not sure how without literally doing ets:tab2list from tests instead of here, but that'd make this one a leaky abstraction 🤔

PR #188, opened 2025-06-04
Upgrade dependencies and runtime versions

NelsonVides:
Support for OTP28.

Also, a test became flaky after the fact that `ets:tab2list/1` might return values in a different order in different versions of OTP, so we here just want to capture the value and compare after sorting.


test/amoc_config_scenario_SUITE.erl
```diff
@@ -207,10 +207,9 @@ update_settings_readonly(_) ->
                         {config_scenario_var1, val1}, %% the same value as set initially
                         {config_scenario_var3, val3}],
     ReadOnlyRet = amoc_config_scenario:update_settings(ScenarioSettings),
-    ?assertEqual({error, readonly_parameters,
```
chrzaszcz, line 210: Could you explain what happened here?

chrzaszcz (commented):
Looks good. Could you just put a short description in the PR and briefly eplain the change in the test suite?

PR #191, opened 2026-02-16
Allow description to be a binary in module attributes parsing

### LINKED ISSUE #190: Allow description to be a binary in module attributes for required variables
When creating a required variable via module attributes:

```elixir
@required_variable %{name: :some_var, description: ~c"some description", default_value: 6}
```

It requires that the description must be a charlist and will error with an `:invalid_attribute` error when it is not.

Allowing description to be a binary would allow easier interop with Elixir.

probably-not:
Closes #190.

I'm by no means an Erlanger so let me know if I need to change anything here!


NelsonVides:
Closing as superseeded by #190, thank you so much for your PR! :)

PR #197, opened 2026-04-27
Introducing setting scope (local/global)

DenysGonchar:
added scope for configuration parameters.
• global parameters can be set on master node only, also they are automatically propagated to the slave nodes.
• local paramteres can be set per node.

also fixed amoc_throttle:stop() for throttles with gradually changing rates

### DIFF
--- a/integration_test/extra_code_paths/path1/dummy_helper.erl
+++ b/integration_test/extra_code_paths/path1/dummy_helper.erl
@@ -33,9 +33,8 @@ test_amoc_dist() ->
         ?assertEqual(N2, N1 + 20, ?comment(Users2)),
         %% remove 10 users
         remove_and_wait(Master, 10),
-        {N3, Max3, Nodes3, _Ids3, Users3} = get_users_info(Slaves),
+        {N3, _Max3, Nodes3, _Ids3, Users3} = get_users_info(Slaves),
         ?assertEqual(N2 - 10, N3, ?comment(Users3)),
-        ?assertEqual(Max2, Max3, ?comment(Users3)),
         ?assertEqual(Nodes2, Nodes3, ?comment(Users3)),
         %% try to remove N3 users
         Ret = remove_and_wait(Master, N3),
@@ -67,6 +66,7 @@ get_users_info(SlaveNodes) ->
     Nodes = lists:usort([Node || {Node, Users} <- Distrib, [] =/= Users]),
     N = length(Ids),
     MaxId = lists:max(Ids),
+    %% logger:error(#{n => N, max_id => MaxId, nodes => Nodes, ids => Ids, distrib => Distrib}),
     {N, MaxId, Nodes, Ids, Distrib}.
 
 add_and_wait(Master, Num) ->
--- a/integration_test/extra_code_paths/path2/dummy_scenario.erl
+++ b/integration_test/extra_code_paths/path2/dummy_scenario.erl
@@ -29,7 +29,15 @@
     #{name => var6, description => "description6"},
     #{name => nodes, description => "this variable is set for docker "
                                     "container via AMOC_NODES env"},
-    #{name => test, description => "this one to be set via REST API"}]).
+    #{name => test1, description => "this variable is to be set via initial settings"},
+    #{name => test2, update => none, description => "this one to be updated"}]).
+
+
+-required_variable([
+    #{name => global_test1, scope => global,
+      description => "this variable is to be set via initial settings"},
+    #{name => global_test2, scope => global, update => none,
+      description => "this one to be updated"}]).
 
 %% parameter verification method
 -export([test_verification_function/1]).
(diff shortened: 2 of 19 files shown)


integration_test/extra_code_paths/path1/dummy_helper.erl
```diff
@@ -67,6 +66,7 @@ get_users_info(SlaveNodes) ->
     Nodes = lists:usort([Node || {Node, Users} <- Distrib, [] =/= Users]),
     N = length(Ids),
     MaxId = lists:max(Ids),
+    logger:error(#{n => N, max_id => MaxId, nodes => Nodes, ids => Ids, distrib => Distrib}),
```
kamilwaz: Could this be a leftover from debugging?

integration_test/extra_code_paths/path2/dummy_scenario.erl
```diff
@@ -29,7 +29,15 @@
...
-    #{name => test, description => "this one to be set via REST API"}]).
+    #{name => test1, description => "this variable is to be set via initial settings"},
+    #{name => test2, update => none, description => "this one to be updated"}]).
+
+
+-required_variable([
+    #{name => global_test1,  scope => global,
+      description => "this variable is to be set via initial settings"},
+    #{name => global_test2,  scope => global, update => none,
+      description => "this one to be updated"}]).
```
kamilwaz:
```suggestion
    #{name => global_test1, scope => global,
      description => "this variable is to be set via initial settings"},
    #{name => global_test2, scope => global, update => none,
      description => "this one to be updated"}]).
```

src/config/amoc_config_scenario.erl
```diff
@@ -123,27 +128,56 @@ verify_settings(UndefinedParameters, Config, [{Name, _} | T]) ->
...
         ReadOnlyParameters ->
             {error, readonly_parameters, ReadOnlyParameters}
     end.
 
+filter_only_changed_params(Config, Settings) ->
+    %% filter out unchanged parameters
+    ChangedParameters =
+        [P || #module_parameter{name = N, value = V} = P <- Config,
+                V =/= proplists:get_value(N, Settings)],
+    {ok, ChangedParameters}.
```
kamilwaz, line 160: Would it make sense to merge this function with `filter_only_updated_parameters/2`? They appear to serve the identical purpose.

test/throttle_SUITE.erl
```diff
@@ -6,65 +6,70 @@
 -compile([export_all, nowarn_export_all]).
 
 -define(DEFAULT_INTERVAL, 60000). %% one minute
--define(RECV(Msg, Timeout), receive Msg -> ok after Timeout -> {error, not_received_yet} end).
+-define(RECV(Msg, Timeout), receive Msg -> ok after (Timeout + 50) -> {error, not_received_yet} end).
```
kamilwaz, line 9: Wouldn't it better to adjust the timeout argument instead of randomly adding 50 ms here?

integration_test/extra_code_paths/path1/dummy_helper.erl
DenysGonchar: indeed it is

test/throttle_SUITE.erl
DenysGonchar: timeouts in test cases are theoretical, but in practice they sometimes fail on slower machines

src/config/amoc_config_scenario.erl
DenysGonchar: the purpose is slightly different, we do not what to run update functions for unchanged parameters. but we want to make a validation that parameters are not read-only and the scope/node pair is correct.

src/config/amoc_config_scenario.erl
```diff
@@ -123,27 +128,56 @@ verify_settings(UndefinedParameters, Config, [{Name, _} | T]) ->
...
+        GlobalParameters ->
+            propagate_global_parameters(GlobalParameters)
+    end;
+%% global parameters cannot be changed on a slave node
+maybe_propagate_global_parameters(_Config, false) -> ok.
+
+propagate_global_parameters(GlobalParameters) ->
+   SlaveNodes = amoc_cluster:slave_nodes(),
+   [amoc_controller:propagate_config(Node, GlobalParameters) || Node <- SlaveNodes],
+   ok.
```
NelsonVides: Indentation here uses 3 spaces, use 4

src/amoc_controller.erl
```diff
@@ -107,6 +108,11 @@ stop_scenario() ->
 update_settings(Settings) ->
     gen_server:call(?SERVER, {update_settings, Settings}).
 
+-spec propagate_config(node(), amoc_config:config()) ->
+    ok | {error, term()}.
+propagate_config(Node, Config) ->
+    gen_server:cast({?SERVER, Node}, {propagate_config, Config}).
```
NelsonVides: `gen_server:cast/2` will only return `ok`, never `{error, term()}`, so either the spec is wrong or the function is misleading. Checking the code, the genserver just does an ets:insert which is pretty instant and can only fail if the table didn't exist, so probably this was never intended to return a tagged error. But, the genserver is explicitly in another node, which might fail, so in that case maybe we want here a `call` to catch cross-node communication errors and propagate the issue.

kamilwaz (approved):

